PR #253, opened 2015-01-04
Add go-collada, a Go package for working with the Collada file format.

shurcooL:
This isn't my package, it's by @GlenKelley, and it's either a fork or simply related to `github.com/go3d/go-collada` by @metaleap. However, it actually works (in my [testing](https://github.com/shurcooL/play/commit/bb12dd54f9d6b69523b695420b4cac9423bddd03#commitcomment-9134222)) and is useful for loading/parsing the Collada file format, unlike the go3d package which [isn't working/finished](https://github.com/go3d/go-collada/issues/1#issuecomment-68077857).

I'm willing to help maintain it if @GlenKelley doesn't want to, but I think adding it to this list is helpful because as far as I know, it's the only (working) Go package for dealing with the Collada file format.


mholt:
Sweet -- good idea!

PR #268, opened 2015-01-27
Compression section and FastLz library.

fromYukki:
Subject


avelino:
Needs at least 3 libs to create a category.

fromYukki:
@avelino ok, which category is more suitable for this library?

avelino:
Places in **utilities** do not have a category specifies for that library.

PR #287, opened 2015-02-21
Add gonum/matrix

lazywei:
Add gonum/matrix, which is the best matrix package I've seen in go currently.
Also, I update the development status of some other matrix related packages.

### DIFF
--- a/README.md
+++ b/README.md
@@ -433,7 +433,7 @@ Join us on IRC at **#awesome-go** on freenode [web access](http://webchat.freeno
 * [plotinum](https://code.google.com/p/plotinum/) - Plotinum provides an API for building and drawing plots in Go.
 * [streamtools](https://github.com/nytlabs/streamtools) - general purpose, graphical tool for dealing with streams of data.
 * [vectormath](https://github.com/spate/vectormath) - Vectormath for Go, an adaptation of the scalar C functions from Sony's Vector Math library, as found in the Bullet-2.79 source code. (currently inactive)
-* [gonum/mat64](https://github.com/gonum/matrix) - The general purpose package for matrix computation. Package mat64 provides basic linear algebra operations for float64 matrices. (currently active)
+* [gonum/mat64](https://github.com/gonum/matrix) - The general purpose package for matrix computation. Package mat64 provides basic linear algebra operations for float64 matrices.
 
 
 ## Serialization
--- a/README.md
+++ b/README.md
@@ -426,13 +426,14 @@ Join us on IRC at **#awesome-go** on freenode [web access](http://webchat.freeno
 * [go-fn](https://code.google.com/p/go-fn/) - Mathematical functions written in Go language, that are not covered by math pkg
 * [gofrac](https://github.com/anschelsc/gofrac) - A (goinstallable) fractions library for go with support for basic arithmetic.
 * [go-gt](https://code.google.com/p/go-gt/) - Graph theory algorithms written in "Go" language
-* [go.matrix](https://github.com/skelterjohn/go.matrix) - linear algebra for go
+* [go.matrix](https://github.com/skelterjohn/go.matrix) - linear algebra for go (has been stalled)
 * [gostat](https://code.google.com/p/gostat/) - A statistics library for the go language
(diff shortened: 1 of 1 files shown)


shurcooL:
How does `gonum/matrix` compare to https://github.com/go-gl/mathgl?

lazywei:
Oh, I wasn't aware of that package. I'll take a look at it later. However, at the very first glance, it seems `mathgl` is more like a specialized package for 3D math computation, while `gonum/matrix` is more like a general purpose linear algebra package. For example, `gonum/matrix` is based on BLAS, so it is very fast for large matrix computation. On the other hand, `mathgl` is specialized for matrix size below 4x4, and though there are variable size matrices/vectors, it seems the author of `mathgl` doesn't encourage developers to use those variable size matrices/vectors.

Anyway, `mathgl` is a very interesting project, and I think I can maybe include it into this PR. How do you think?

shurcooL:
Thanks for elaborating. It does sound like they serve slightly different needs, `mathgl` being primarily for graphics where matrices of size 4x4 or smaller are most common.

`mathgl` is already on the Awesome Go list, there's no need to add it a second time. It's listed under the [OpenGL](https://github.com/avelino/awesome-go#opengl) section.

lazywei:
@shurcooL got it, thanks for explanation. I have no other comment for this PR then.
Thanks.

README.md
```diff
@@ -426,13 +426,14 @@ Join us on IRC at **#awesome-go** on freenode [web access](http://webchat.freeno
...
 * [goraph](https://github.com/gyuho/goraph) - A pure Go graph theory library(data structure, algorith visualization)
 * [mudlark-go](https://code.google.com/p/mudlark-go-pkgs/) - A collection of packages providing (hopefully) useful code for use in software using Google's Go programming language.
 * [plotinum](https://code.google.com/p/plotinum/) - Plotinum provides an API for building and drawing plots in Go.
 * [streamtools](https://github.com/nytlabs/streamtools) - general purpose, graphical tool for dealing with streams of data.
-* [vectormath](https://github.com/spate/vectormath) - Vectormath for Go, an adaptation of the scalar C functions from Sony's Vector Math library, as found in the Bullet-2.79 source code.
+* [vectormath](https://github.com/spate/vectormath) - Vectormath for Go, an adaptation of the scalar C functions from Sony's Vector Math library, as found in the Bullet-2.79 source code. (currently inactive)
+* [gonum/mat64](https://github.com/gonum/matrix) - The general purpose package for matrix computation. Package mat64 provides basic linear algebra operations for float64 matrices. (currently active)
```
shurcooL: Leave the ` (currently active)` part out. It's the default. There's no need to state it explicitly.

README.md
```diff
@@ -426,13 +426,14 @@ Join us on IRC at **#awesome-go** on freenode [web access](http://webchat.freeno
 * [go-fn](https://code.google.com/p/go-fn/) - Mathematical functions written in Go language, that are not covered by math pkg
 * [gofrac](https://github.com/anschelsc/gofrac) - A (goinstallable) fractions library for go with support for basic arithmetic.
 * [go-gt](https://code.google.com/p/go-gt/) - Graph theory algorithms written in "Go" language
-* [go.matrix](https://github.com/skelterjohn/go.matrix) - linear algebra for go
+* [go.matrix](https://github.com/skelterjohn/go.matrix) - linear algebra for go (has been stalled)
```
shurcooL: If/when this PR gets merged, I'm planning to make another PR that removes the stale/inactive repos, espcially since there's a new/active one that is said to be better (can anyone confirm?). I don't think we want to keep such repos according to the [quality standard](https://github.com/avelino/awesome-go/blob/master/CONTRIBUTING.md#quality-standard):

> Actively maintained (even if that just means acknowledging open issues when they arise)

lazywei:
I just pushed a new commit to remove the redundant status description. :smile:

README.md
mholt: That is probably a good idea. Don't forget the vectormath one below. At least the gonum repo looks pretty solid and still active.

I'll merge this PR now though.

mholt:
(TODO: Remove the two inactive repos from this change.)

PR #291, opened 2015-02-28
Update README.md

mgutz:
Refactored build tools to their own section
Added some of my libraries


mholt:
That ansi package might be more suitable under CUI/Console tools... that's the only suggestion I have before merging this. Thanks!

mholt:
Ping @mgutz :smile:

Issue #282, opened 2015-02-08
CUI vs. CLI


netpoetica:
I noticed that [the section regarding tools for building Command Line Interfaces](https://github.com/avelino/awesome-go#cui) is titled "CUI", which was brought to my attention by [a PR to awsome-cpp](https://github.com/fffaraz/awesome-cpp/pull/71). However, a quick Google search for "CUI" or "CUI Applications" yields very few results for this being a thing. Is it common to refer to CLI as CUI in the Go community? Or are people trying to make it a thing?

If you're open to a pull request, I can change the heading from "CUI" to "CLI", which is, as far as my common knowledge and Google search abilities show, a much more well-understood name for this.

shurcooL:
I think CUI is a typo, it should be CLI. There's nothing Go-specific about it as far as I know. I've never heard of CUI either.

That PR would be welcome (unless someone else chimes in and explains the situation).

shurcooL:
Okay, it might not be a typo, but it's certainly a lot less common than CLI, so I think we should change it to CLI.

See #58 for origin of the term in this repo. Even in that PR, it was unclear what CUI stood for.

mholt:
It's not a typo.

CUI = Console User Interface. It's a kind of GUI that runs inside the terminal, like curses or, for example, this: https://github.com/buger/goterm or (this, in Node.js: https://github.com/yaronn/blessed-contrib). Sometimes also called a Text User Interface (TUI) but that's even less common. (top and htop also count as CUI apps, if you've ever used those.)

CLI = Command Line Interface. Less of a GUI and more what you're used to on the command line, just line-by-line text input/output.

I think we should leave CUI its own category.

shurcooL:
Ah, my bad. I missed that these were indeed libraries for text/console UIs, not command line UIs.

In that case, CUI is definitely more appropriate and should not be changed to CLI, since that would be incorrect.

netpoetica:
I see. CUIs in this sense still are CLI applications though - it's a bit like a square/rectangle thing. I don't want to try to make this an argument of semantics, but essentially the idea of "CUI" is basically "libraries for advanced command line interfaces." 

[...]

PR #292, opened 2015-02-28
divide CUI into std and advanced Command Line

netpoetica:
https://github.com/avelino/awesome-go/issues/282

Notes: It was not immediately clear whether I should move the command line email tool to Standard Command Line or leave as a utility (https://github.com/sanbornm/mp). I think the best option is to leave it as a utility because it isn't a tool for building command line apps, it is just a command line app itself.


mholt:
Well, after seeing some of the prior discussion on how to classify console-related packages, I think this change is a step in the right direction. Thank you!

PR #303, opened 2015-03-06
Add decimal

vadimg:


avelino:
Needs more library to create a category, at least 3.
your library doesn't fit in "Science and Data Analysis"?

mholt:
Hmm, the category limit makes sense. But I do agree that it is first and foremost a financial package. (Most science that I have seen, like simulations and such, can live with minor floating point errors, and the space savings is valuable enough that they just use floats instead of a custom type.)

It's a good package though so let's find a spot for it. :+1: I could see it going in Science and Data Analysis... at least for now.

avelino:
It's not a rules have 3 and yes a good sense, even if rule fose (every rule has an exception).

If you think worth having this new 125cc category, that's fine by me, remembering that the awesome-go is not for me but for us!

vadimg:
The reason I created a "Financial" category is that a lot of software development is financial-related, and I was surprised that go didn't have any financial packages - yet.

I can put it in the "Science and Data Analysis" though - it doesn't really fit in there but it arguably also doesn't fit into the financial section since it's just a type, which itself has nothing to do with money.

mholt:
Tell ya what, maybe having a financial category will encourage more emphasis on good financial libs. As-is, the ordering needs to be fixed so it's in alphabetical order anyway. So fix that, and if you want, go ahead and keep it in a Finance category. I think we'll take it either way! Just make sure it's alphabetical. :+1:

vadimg:
Ok, I put it back into the financial category :)

mholt:
Voila. Thanks!

avelino:
+1 :D

PR #306, opened 2015-03-10
Rename geo_pattern to geopattern

pravj:
> Fix lexicographical ordering of other packages in Images section too.

Renamed it after this discussion, http://git.io/geopattern-naming-discussion


shurcooL:
LGTM.

PR #318, opened 2015-03-26
Add gh, package for GitHub-centric ops

rjeczalik:


mholt:
Is this software or a code library? We might want to move it to a more appropriate category first.

rjeczalik:
Both, but more library than software. "Third-party APIs" category is mostly for clients, and this package implements a server for webhooks. I'm not sure where it belongs, open to suggestions.

mholt:
What about Version Control? It appears to be directly related to GitHub and git events. If you feel like that's an appropriate category, feel free to make the change and we'll merge it in. If not, just do what you feel is best. (Even if it's Misc.)

rjeczalik:
Sure, moved to Version Control.

PR #324, opened 2015-03-29
Add go-steam, library to interact with Steam servers

kidoman:
Ref: https://developer.valvesoftware.com/wiki/Source_RCON_Protocol


shurcooL:
The linked repo has an import path of `github.com/sostronk/go-steam`, but the readme says:

```
go get github.com/kidoman/go-steam
```

Which is the right import path?

norwack:
@shurcooL The first one was updated in february, the latter was last updated in december. "sostronk" is an organization with only @kidoman as a member, so i'd assume `github.com/sostronk/go-steam` is the correct import path.

It's also 8 commits ahead of kidoman:master.

shurcooL:
In that case, it's README should be updated to match.

kidoman:
Will do that!

shurcooL:
LGTM.

dukex:
The project has a documentation?

shurcooL:
Its godoc is at https://godoc.org/github.com/sostronk/go-steam.

PR #330, opened 2015-04-03
Added shell2http

msoap:
shell2http  - executing shell commands via http server


norwack:
This should be under [this category](https://github.com/avelino/awesome-go#other-software)

msoap:
Fixed in https://github.com/msoap/awesome-go/commit/26a3542cb37db2757996db035236a391aeb5ba5e

PR #333, opened 2015-04-06
Added two projects

xyproto:
Added two projects:
* [permissions2](https://github.com/xyproto/permissions2)
* [algernon](https://github.com/xyproto/algernon)


mholt:
Algernon appears to be full software rather than just a library/package. It should probably be moved somewhere under Tools/Software Packages.

xyproto:
I agree. Made the suggested changes.

mholt:
Can we put it in Other Software instead of its own category? (Unless you feel it would be more fitting in DevOps)

xyproto:
A webserver is not a library and not a tool, though. How about a category for server applications?

mholt:
Hmm. I admit "Tools" isn't the best name for that category. @shurcooL What if we refactor that into a "Software" category for Go software projects that aren't intended for use as stand-alone libraries?

xyproto:
Could something like this work? Created a sub-category for server applications.

shurcooL:
Looks okay to me.

mholt:
Okay, we'll start with that. Thanks!

xyproto:
Thanks for merging and thanks for maintaining awsome-go!

PR #339, opened 2015-04-10
Add taglib

wtolson:


mholt:
This package doesn't appear to be active (also FYI the godoc link is old).

wtolson:
Good find, updated the godoc link. Still active, just low volume.

mholt:
Okay, thanks!

PR #360, opened 2015-04-22
Added ServeMux to the list of Web Frameworks

hallas:
Hi :smile: 

Please let me know if you have any questions.

Best regards

### DIFF
--- a/README.md
+++ b/README.md
@@ -709,6 +709,7 @@ Join us on IRC at **#awesome-go** on freenode [web access](http://webchat.freeno
 * [Resoursea](https://github.com/resoursea/api) - A REST framework for quickly writing resource based services.
 * [Revel](https://github.com/revel/revel) - A high-productivity web framework for the Go language.
 * [sawsij](http://sawsij.com/) - lightweight, open-source web framework for building high-performance, data-driven web applications.
+* [ServeMux](http://servemux.com) - Idiomatic Go http.Handler framework.
 * [tango](https://github.com/lunny/tango) - Micro & pluggable web framework for Go.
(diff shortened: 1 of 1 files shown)


shurcooL:
From the Simple Example on that page,

> Keep in mind that even though the example above uses the servemux/common package

The example does not appear to use the servemux/common package.

```Go
import (
	"github.com/servemux/app"
	"github.com/servemux/buffer"
	"github.com/servemux/responsetime"
	"io"
	"net/http"
)
```

Is that a mistake?

mholt:
For this and the one above it (somehow missed that one - I can fix that), it should link to the GitHub page.

shurcooL:
> it should link to the GitHub page.

Why? Shouldn't it link to the _canonical home page_ of the Go package, whatever that might be, including a github repo, a bitbucket repo, a godoc.org link, or a dedicated website as in this case?

mholt:
For consistency I suppose. But we can keep it as-is.

shurcooL:
I think it's consistent in that each link points to the canonical home page. The package authors can decide what they like as their home page and if they want to use a GitHub page with readme like most others or otherwise.

hallas:
Thanks for noticing the mishap in the text for the example.

I used the homepage because I saw someone else had done the same in the list. I am fine with whichever standard you guys decide. In my case, the homepage represents ServeMux better because it is a collection of packages.

hallas:
Updated the copy on the website. Again, thanks for catching that.

shurcooL:
> In my case, the homepage represents ServeMux better because it is a collection of packages all in the same category and written in the same style and made to work together.

Yep, that's why I wanted you to be feel free to use what best represents your project.

> Again, thanks for catching that.

No problem, thanks for fixing it quickly!

PR #362, opened 2015-04-23
Adding the golang-set package which offers thread-safe and non-thread-sa...

deckarep:
Hello,

This library offers unit-tested, benchmarked high-performance set datastructures for Go.  It comes in both thread-safe and non-threadsafe versions and implements a common interface between both of them.

https://github.com/deckarep/golang-set

Note: I added this under a new category of data-structres since I didn't think it really belonged anywhere else.

Please consider adding it!

-Thanks!


mholt:
I agree the new category is merited. The Database category has a pseudo-section for data structures. Would you mind moving all those into the new Data Structures category? That'd help us make this a lot better, I think.

deckarep:
Will do!

deckarep:
@mholt - this should be good to go now!  Thanks for looking at this.

mholt:
Sweet, we just gotta take care of that merge conflict - could you do that? Then I'll merge it in.

deckarep:
@mholt - ok, should be good to go.

mholt:
Thank you!

Issue #364, opened 2015-04-23
test duplicate link


avelino:
Today the tests tests whether this in alphabetically, it would be important to check if there is already a link (ex: github.com/...)

PR #366, opened 2015-04-23
Add Test to catch duplicated links

dukex:
Related #364


dukex:
We have some duplicated hash links

/cc @avelino @shurcooL @mholt

PR #370, opened 2015-04-24
Added some libraries

henryx:


dukex:
Hi @henryx do you can fix the build?

PR #375, opened 2015-04-27
add logrus as a standalone lib

mariokostelac:
It seems reasonable (from my point of view) to have logrus as a standalone library. If you do not know what is logrus and do not care about a plugin for something (which is logrusly), you will miss logrus itself.


avelino:
Fork lib?
Official: https://github.com/Sirupsen/logrus

mariokostelac:
Sorry, pasted wrong link.

PR #392, opened 2015-05-11
Add cli by mitchellh

pichuio:


README.md
```diff
@@ -113,6 +113,7 @@ Join us on IRC at **#awesome-go** on freenode [web access](http://webchat.freeno
 *Libraries for building standard or basic Command Line applications*
 
 * [cli](https://github.com/codegangsta/cli) - A small package for building command line apps in Go.
+* [cli](https://github.com/mitchellh/cli) - A Go library for implementing command-line interfaces.
```
mholt: Since these two adjacent libraries have the exact same name, maybe it would be best to include their owner username, for example, `codegangsta/cli` and `mitchellh/cli` for clarity.

pichuio:
i think that is a good idea, i will change it

mholt:
Thanks! But that broke the alphabetical ordering. If you could just switch those around a bit then we'll merge it in. :smile:

pichuio:
now it should work :smile:

PR #396, opened 2015-05-13
Add albrow/zoom

albrow:
Though it is not exactly an ORM, it does function similarly. I see other ORM-like libraries in this category so I believe it is the appropriate place. Happy to move it somewhere else if you think it would be more appropriate.

Cheers!


shurcooL:
LGTM. zoom is an active, well documented Go package.

PR #401, opened 2015-05-16
Update import path of github_flavored_markdown package.

shurcooL:
It has moved out into a standalone repo. See shurcooL/go#19 for rationale.


shurcooL:
LGTM, merging because I'm the author of the package and I can confirm this is its new import path. :)

PR #408, opened 2015-05-18
Add ryszard/goskiplist.

rscnt:
A better implementation of skiplist in go.


rscnt:
position of the ryszard/goskiplist repo has changed. @shurcooL

PR #424, opened 2015-05-23
Add TARGA image format decoder/encoder package.

shurcooL:
There is no other tga encoder/decoder package in the awesome-go list, so I think it would be beneficial to add it. This is the only and best tga package I know of.

> Supports RLE and raw TARGA images with 8/15/16/24/32 bits per pixel, monochrome, truecolor and colormapped images. Correctly handles origins and attribute type in extensions area. Passes TGA 2.0 conformance suite (http://googlesites.inequation.org/tgautilities).

I've opened https://github.com/ftrvxmtrx/tga/issues/4 to see if it's still well maintained, since it was last updated 3 years ago. But it's complete and functional, so I wouldn't hold it too much against it.

/cc @ftrvxmtrx


shurcooL:
ftrvxmtrx/tga#4 has been resolved very rapidly, so that should address any concerns about it being unmaintained.

PR #434, opened 2015-05-28
Added diegobernardes/ttlcache under DataStructures

diegobernardes:
I'm the author of this package. 
The package is with 100% test coverage.


mholt:
Looks good, but why the fork and not the original repo?

diegobernardes:
TTLCache was forked from [wunderlist/ttlcache](https://github.com/wunderlist/ttlcache) to add extra functions not avaiable in the original scope.
The main differences are:

1. A item can store any kind of object, previously, only strings could be saved
2. There is a option to add a callback to get key expiration
3. The expiration can be either global or per item
4. Can exist items without expiration time
5. Expirations and callbacks are realtime. Don't have a pooling time to check anymore, now it's done with a heap.

PR #456, opened 2015-06-21
Add go-yara

hillu:


mholt:
I wonder if this would be better situated in the Security category. (I realize it's 'text processing' too, but a very limited subset of it.) What do you think?

hillu:
* Matt Holt:

> I wonder if this would be better situated in the Security category. (I
> realize it's 'text processing' too, but a very limited subset of it.)
> What do you think?

I agree Security is probably a better match since yara is more about
pattern matching than text processing.

Cheers,
-Hilko

hillu:
Moved the go-yara entry to the Security section

mholt:
Thanks!

PR #459, opened 2015-06-23
Added vdobber's chart

arranubels:


mholt:
Your fork is almost 600 commits behind master. Yikes! Sync up with the latest then update your PR ;)

PR #468, opened 2015-06-29
Added go-outdated

firstrow:


shurcooL:
This looks related to [`gostatus`](https://github.com/shurcooL/gostatus), which displays Go packages that have updates with a `+` symbol. Nothing wrong with that, just making an observation.

firstrow:
True. I think if go-outdated has nothing new, we should not add it to the list.

matrixik:
Hi, I think it's nice addition. Added in https://github.com/avelino/awesome-go/commit/cda82d345671c89b2d022f99d4ad55e77dd45b30 Thank you.

PR #471, opened 2015-07-01
Add gofpdf library and documents header to list

jung-kurt:


mholt:
I'm good with adding this package -- it's very well documented. (Good work!) Is there an existing category we could use? I don't know if it qualifies as Text Processing -- what would you say is the closest existing category?

If none are good enough, I suppose we can make a new one. Or morph a similar one so it includes "File Formats" or "Documents" or something.

jung-kurt:
>
> I'm good with adding this package -- it's very well documented.


Thanks!

Is there an existing category we could use? I don't know if it qualifies as
> Text Processing -- what would you say is the closest existing category? If
> none are good enough, I suppose we can make a new one. Or morph a similar
> one so it includes "File Formats" or "Documents" or something.


Maybe change *Text Processing* to *Text and Document Processing*?

-- Kurt

cassiobotaro:
gofpdf is already in category Template Engines.

PR #478, opened 2015-07-04
Adding delts to Testing > Fuzzing and delta-debugging/reducing/shrinking

jasonmoo:


jasonmoo:
Tests are written as Examples, they will fail if the Stdout output does not match.

cassiobotaro:
@jasonmoo  For  quality standards please follow:  https://github.com/avelino/awesome-go/blob/master/.github/PULL_REQUEST_TEMPLATE.md

PR #482, opened 2015-07-04
Adding eq to Data Structures

jasonmoo:


shurcooL:
There are only 3 exported identifiers in this package, and all 3 of them are undocumented:

```
eq.go:6:2: exported type EQ should have comment or be unexported
eq.go:14:1: exported function NewEQ should have comment or be unexported
eq.go:70:1: exported method EQ.Close should have comment or be unexported
```

If you resolve that, this LGTM.

Also, it'd be helpful to add a Go package summary. You already have it written in the README file, so it's just a matter of adding the following comment header to the top of your .go file, then it'll show up in godoc:

```Go
// Package eq offers a simple queue that expands via container/list and uses channels
// for Enqueue and Dequeue actions.
//
// Allows enqueue/dequeue via channels that will never block.
package eq
```

There is a helpful article on golang.org on [this topic](https://blog.golang.org/godoc-documenting-go-code). There's also a section on [commentary](http://golang.org/doc/effective_go.html#commentary) in Effective Go.

I point this out because you've submitted a large number Go packages recently and they seem to lack proper Go documentation.

matrixik:
Thank you for contribution but so small package and still no docs in code? Closing.

PR #484, opened 2015-07-05
Clarifications to quality standards

mholt:
Proposed clarifications to the quality standards, to help avoid confusion.

The previous wording has existed since I put something together real quick before the project became widely known; it's time to be a little clearer about what is expected. I also changed some ordering that seems to flow better as you read it.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -20,16 +20,18 @@ Please contribute links to packages/projects you have used or are familiar with.
 
 ## Quality standard
 
-To stay on the list, package repositories should adhere to these quality standards:
+To be on the list, project repositories should adhere to these quality standards:
 
-- Generally useful to the community
-- Functional
-- Actively maintained (even if that just means acknowledging open issues when they arise)
-- Stable, or progressing toward stable
-- Documented (preferably godoc.org)
-- Tests are preferred (when possible)
+- Working
+- Generally useful to the wider community of Go programmers
+- Actively maintained
+  - Regular, recent commits
+  - For finished projects, issues and pull requests are responded to
+- Stableor progressing toward stable
+- Thoroughly documented (README, godoc comments, etc.)
+- Tests, where practical
 
 
 ## Reporting issues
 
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -20,16 +20,18 @@ Please contribute links to packages/projects you have used or are familiar with.
...
-- Documented (preferably godoc.org)
-- Tests are preferred (when possible)
+- Code is functional
+- Generally useful to the wider community of Go programmers
+- Actively maintained
+  - Regular, recent commits
+  - Or, for finished projects, issues and pull requests are responded to
+- Stable or progressing toward stable
+- Thoroughly documented (README, godoc comments, etc.)
+- Tests, where practical
```
shurcooL: Suggestion. I think we should enforce it to follow [recommended Go style](https://github.com/golang/go/wiki/CodeReviewComments) (not necessarily perfectly, but to a large degree) and therefore have no or little `golint` issues.

The only exceptions should be existing packages that are extremely useful, popular and well known (but those tend to follow recommended Go style anyway).

I think high quality Go packages tend to already do this anyway, and this extra effort will filter out many unhelpful packages where authors did not put in a lot of time and effort or care into making the Go package as nice as it can be.

mholt: I like that idea, although I do have some reservations about golint. It's a great tool but the output is mainly suggestions... for example, I disagree with it on several points of style.

What about goreportcard.com - it does use golint but it also incorporates other factors into the grade...

shurcooL:
Another point, I think for PRs that propose to remove something (because it's stale, out of date, not helpful, abandoned, etc.), let's adopt a policy they should be LGTM'ed and merged no sooner than 7 days after being open.

That way there's a window for whoever is affected to chime in and comment ("oh, it's temporarily down, will be fixed tomorrow", etc.). 7 days should be enough time for them to likely see it. Merging a PR that removes something 2 hours after it's created is not enough time, and can cause misunderstandings.

mholt:
That's a good idea.

CONTRIBUTING.md
shurcooL: Using goreportcard.com sounds fair. It includes lint, vet and a few others in one, which is great.

However, I see there's a yellow notice on top that says "Go Report Card shutting down". http://goreportcard.com/shutdown/

PR #488, opened 2015-07-07
Add boomfilters and go-datastructures packages

tylertreat:


avelino:
does the PR in 1 commit please.

PR #492, opened 2015-07-11
Adding in Buffstreams - Streaming protobuffs message data over TCP ma…

StabbyCutyou:
…de easy

One of the guidelines states tests are optional, but encouraged.

I don't have a full proper set of tests, but I do have some basic tests and a test client / server that are used to run benchmarks.

### DIFF
--- a/README.md
+++ b/README.md
@@ -474,6 +474,7 @@ See [go-hardware](https://github.com/rakyll/go-hardware) for a comprehensive lis
 
 *Libraries for working with various layers of the network*
 
+* [buffstreams](https://github.com/stabbycutyou/buffstreams) - Streaming protocolbuffer data over TCP made easy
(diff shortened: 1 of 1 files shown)


StabbyCutyou:
Just as an update, I'm nearly done with a branch that adds in some proper go tests, so that should cover any issues around lack of coverage.

This PR will also bring the "interface" of the library much closer to how the internal golang socket code looks, which should make working with it more natural for go devs.

StabbyCutyou:
As a final update, I've got the start of proper tests along with a test client/server, and was able to push my benchmarks to 1.1M messages per second on a single listener. It also now looks very very close to the underlying go libraries that it wraps, making it a little more familiar to work with.

mholt:
Good work, thanks :smile:

PR #496, opened 2015-07-15
Added Machinery to Messaging section.

RichardKnop:
Machinery is an asynchronous task queue/job queue based on distributed
message passing. Think of it as Celery for Golang.

So called tasks (or jobs if you like) are executed concurrently either
by many workers on many servers or multiple worker processes on a single
server using Golang's goroutines.

See the repository here: https://github.com/RichardKnop/machinery

Hacker News post: https://news.ycombinator.com/item?id=9520712


mholt:
Oh noez, out of order makes me said :cry:

RichardKnop:
@mholt Done :)

PR #498, opened 2015-07-21
goraft is unmaintained

aviau:
> UNMAINTAINED: A Go implementation of the Raft distributed consensus protocol.

### DIFF
--- a/README.md
+++ b/README.md
@@ -261,7 +261,6 @@ Please take a quick gander at the [contribution guidelines](https://github.com/a
 
 *Packages that help with building Distributed Systems.*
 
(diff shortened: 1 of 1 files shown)


shurcooL:
From https://github.com/goraft/raft/blob/master/README.md#overview:

> **NOTE**: This project is unmaintained. If you are using goraft in a project
> and want to carry the project forward please file an issue with your ideas and
> intentions. The original project authors have created new raft implementations
> now used in [etcd](https://godoc.org/github.com/coreos/etcd/raft) and [InfluxDB](https://godoc.org/github.com/influxdb/influxdb/raft).

LGTM.

We could add one of the new implementations later on.

PR #503, opened 2015-07-23
Add poorny/translate package

marioidival:


avelino:
Added in ` Third-party APIs`!

PR #505, opened 2015-07-26
govalidator belongs in Validation section

mrfuxi:
Since there is a Validations section `asaskevich/govalidator` fits there better then under "Text Processing/Utility"


shurcooL:
LGTM.

PR #508, opened 2015-07-29
Add sparsevector

philpearl:
I've not found another sparse vector library in Go, so here's mine in case someone else finds it useful.


philpearl:
I'd say not.  It's more maths - linear algebra, belonging close to the other matrix things in this section.

philpearl:
@alaska Any further thoughts?

PR #509, opened 2015-07-30
Adding go-nmea to Utilities

adrianmo:


shurcooL:
The library looks good, but not sure if Utilities is the best section for it. Did you not find anything more appropriate?

It feels like it should go together with other file format parsers somewhere. Perhaps Text Processing?

adrianmo:
Yes, you are right, probably "Test Processing" more appropriate. Let me move it there.

adrianmo:
Done

shurcooL:
LGTM.

Issue #497, opened 2015-07-16
gb build tool


fzerorubigd:
this one https://github.com/constabulary/gb is a great tool.
I don't know which section is the best part for this one, so I just open an issue instead of PR.

shurcooL:
gb (/cc @davecheney) is indeed an awesome tool. It's one of a kind, being a replacement for the `go` tool.

I would imagine the best place for it is under the [Tools](https://github.com/avelino/awesome-go#tools) category, but what subcategory is harder to say... Clearly, there's nothing specific for it now. I think [Code Analysis](https://github.com/avelino/awesome-go#code-analysis) category is the best fit, given that it contains other tools related to working with Go code like `goimports`, `lint`, etc. It seems close enough so we don't need to create a separate subcategory. Or we could create "Build Tools" and gb would be the only entry.

What do others think?

alaska:
I don't think Code Analysis is quite the right fit. My vote would be to keep that category for static analysis tools. If you have to create a new category, I'd suggest something like "go project tools", "go build tools" or just put it directly under the tools section.

shurcooL:
> put it directly under the tools section.

I kinda like this. I don't think there will be many viable alternatives to `go` tool, and `gb` can be a standalone entry under Tools header until there's a better category to move it into.

Unless there are any objections? (Disclosure, I know @davecheney personally and I'm familiar with `gb`... But I suppose this applies to many people hehe. :) )

PR #514, opened 2015-08-02
Add `colorgo`, a `go build` colorizer

songgao:


shurcooL:
Hi @songgao,

This looks like a cool tool!

However, I think it's in the wrong section. It's not a CLI library, but a tool for working with go tool. I think it should be moved somewhere under the [Tools](https://github.com/avelino/awesome-go#tools) section.

songgao:
Make sense. I'll `git --amend` it then

songgao:
@shurcooL Actually on a second thought, what do you think of the [Utilities](https://github.com/avelino/awesome-go#utilities) section?

shurcooL:
Utilities is better than your original section, but I still think Tools section is better. Utilities contains general utilities for doing various things. Tools contains utilities specifically for working with Go.

songgao:
@shurcooL Thanks for the explanation. I amended the commit to move it to Tools section. (I assumed you meant the [DevOps Tools](https://github.com/avelino/awesome-go#devops-tools) subsection)

shurcooL:
I kinda wanted to see it under Code Analysis, but it's not exactly Code Analysis. I personally would want to see a similar section added there. But I suppose DevOps Tools will do too. I don't know about others, but DevOps means something else to me (and it seems to have quite a few things that IMO aren't ideal there).

If you're okay with it, LGTM, but I think you could also make a new subsection under Tools where your tool could go.

shurcooL:
See https://github.com/avelino/awesome-go/issues/497 for a related discussion. I feel like we should add another section next to Code Analysis to fill the gap.

songgao:
In fact I didn't think it was exactly a `DevOps Tool` either :D 

I like the idea of adding another subsection. How about "Build Tool" or "Coding Assistance"?

shurcooL:
Between those two, I like "Build Tools" better.

However, I'd want to move tools like `gb,` `goimports`, `lint` and you tool in there. So it's not just build tools, but rather tools related to Go.

So how about "Go Tools" as the new subsection name?

songgao:
Sounds good to me! I also updated this PR; I guess moving other packages should be in a separate PR?

shurcooL:
Yep.

LGTM. Thanks!

songgao:
Sure thing. Thanks for the suggestions!

dukex:
@songgao You need put this new section in section index

songgao:
@dukex Done! Sorry for overlooking that earlier

shurcooL:
@dukex Thanks for catching that, I overlooked it.

shurcooL:
LGTM.

dukex:
Thanks @songgao and @shurcooL

songgao:
Thanks for merging :D

PR #516, opened 2015-08-03
Add timeutil.

leekchan:
I added timeutil.

Timeutil is a set of useful extensions (Timedelta, Strftime, ...) to the golang's time package.

Thanks :)


shurcooL:
LGTM.

PR #526, opened 2015-08-09
add go-stun

ccding:


shurcooL:
Whoa, this is really cool and implements an RFC spec! Thank you!

LGTM.

PR #527, opened 2015-08-09
Add "yakvs"

sci4me:


shurcooL:
Given that it's a key value store library, I think it'd helpful if it had docs. Right now it appears to have none. https://godoc.org/github.com/sci4me/yakvs/yakvs

shurcooL:
Oh, is this meant to be a command rather than a library? If so, my comment above may be less applicable.

sci4me:
Yes it is not meant to be a library at all, it is just a server. The protocol describes the TCP "api".

sci4me:
I have added docs, but please, if you think they need improvement, let me know... I am not a skilled documentation writer.

shurcooL:
Looks great, thanks for doing that!

LGTM.

PR #528, opened 2015-08-09
add wellington and low level libsass wrapper

drewwells:
Adds the go-libsass lib for writing your own Sass tooling. The default
implementation, wellington, implements a tool for working on Sass
projects. Wellington adds the same spriting functions as available in Compass.

Signed-off-by: Drew Wells <drew.wells00@gmail.com>

### DIFF
--- a/README.md
+++ b/README.md
@@ -173,6 +173,7 @@ Please take a quick gander at the [contribution guidelines](https://github.com/a
 
 * [c6](https://github.com/c9s/c6) - High performance SASS compatible-implementation compiler written in Go
 * [gcss](https://github.com/yosssi/gcss) - Pure Go CSS Preprocessor.
+* [go-libsass](https://github.com/wellington/go-libsass) - Go wrapper to the 100% Sass compatible libsass project.
 
 ## Data Structures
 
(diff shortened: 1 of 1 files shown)


drewwells:
This has been sitting here for a while. Is there something you need me to change for acceptance?

shurcooL:
No, it probably just slipped through because this project has a lot of traffic and not enough time from maintainers.

Looked it over, LGTM.

Although I had a hard time figuring out if `wellington` is a Go library or a binary tool. If it's the latter, it should go under another section. But I _think_ it's a library. Can you confirm?

shurcooL:
Hmm, looking closer, I think it is primarily a binary after all. https://godoc.org/github.com/wellington/wellington is imported by [only one package](https://godoc.org/github.com/wellington/wellington?importers), `wt`, which is the binary.

Since this section is for Go library packages only, you should move `wellington` to somewhere under [Tools section](https://github.com/avelino/awesome-go#tools) please?

drewwells:
wellington is more dedicated to the binary, the wrapper is more general purpose.

drewwells:
moved it to other tools, it didn't appear to fit into any of the other tools categories. Thanks!

shurcooL:
LGTM, thanks!

PR #529, opened 2015-08-09
add evaler

soniah:


shurcooL:
LGTM.

This is a little like Rob Pike's ivy, at least in spirit, right?

soniah:
Hmmm, yes. His is obviously way more ambitious :)

PR #542, opened 2015-08-16
Named anchor broken when not on GitHub

anacrolix:
I imagine there are others to do. On GitHub the named anchor scheme works, only because GH inserts those anchors for you. On http://awesome-go.com/, this doesn't work for headers that contain multiple words.

### DIFF
--- a/README.md
+++ b/README.md
@@ -95,7 +95,7 @@ Please take a quick gander at the [contribution guidelines](https://github.com/a
 * [waveform](https://github.com/mdlayher/waveform) - Go package capable of generating waveform images from audio streams.
 
 
-## Authentication & OAuth
+## <a name="authentication--oauth"></a>Authentication & OAuth
 
 *Libraries for implementing authentications schemes.*
 


dgryski:
Yes, there are a few other broken links.  We should probably just be consistent and make *everything* an explicitly named anchor.

shurcooL:
How is http://awesome-go.com/ generated (the HTML from Markdown)? If we use [`github_flavored_markdown`](https://godoc.org/github.com/shurcooL/github_flavored_markdown) package, it generates the anchors... Just like GitHub.

shurcooL:
I see, it's done in frontend via GitHub API.

https://github.com/avelino/awesome-go/blob/gh-pages/assets/awesome-go.js

However, there's a better API endpoint to use. Instead of fetching the raw .md content and converting it via `Showdown.converter`, we can just use GitHub's Markdown rendering endpoint:

https://developer.github.com/v3/markdown/

dukex:
I think uses the markdown endpoint is a better approach for fix it.

avelino:
Pls read and resend PR: https://github.com/avelino/awesome-go/blob/master/CONTRIBUTING.md

PR #555, opened 2015-08-25
add beego config

hvnsweeting:
to me it's even better than viper for config reading. 
If you use viper.GetInt("foo") when ``foo: bar``, you will get int 0.
Beego config returns error.

### DIFF
--- a/README.md
+++ b/README.md
@@ -146,6 +146,7 @@ Please take a quick gander at the [contribution guidelines](https://github.com/a
 
 *Libraries for configuration parsing*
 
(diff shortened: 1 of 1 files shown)


shurcooL:
Beego is [already added](https://github.com/avelino/awesome-go#web-frameworks) under Web Frameworks. Do you think it makes sense to add more of its individual packages separately?

shurcooL:
There's also an entry for beego's orm package.

hvnsweeting:
yes, seem beego included many good things though it's not clearly for non-beego user know to use it as author put everything in 1 repo. I'm closing.

PR #557, opened 2015-08-25
Add machine learning package - goml

mrfuxi:
As in title. Lib has set of algorithms to use


alaska:
Considering the overlap with https://github.com/josephmisiti/awesome-machine-learning#go, I would think you'd either want to pull in all of the projects listed in that source or simply link there.  Thoughts?

mrfuxi:
Nice find! I would also keep the list in once place. Question is where

dukex:
For me it's ok keep listed here and there.

@avelino @mholt @shurcooL

mholt:
LGTM.

avelino:
LGTM

alaska:
So long as you keep them in sync. Or at least do a one-time import of the entire section instead of just goml

dgryski:
Could we just link to it like we do in the hardware section?

shurcooL:
> Could we just link to it like we do in the hardware section?

That sounds like a much better way. Why spend effort keeping them in sync manually?

PR #558, opened 2015-08-27
Add github.com/vectaport/flowgraph to Distributed Systems

vectaport:
The flowgraph package provides a mechanism to augment goroutines and channels with a ready-send MPI-style protocol that gives the developer tools to manage deadlock and congestion, while ensuring low-latency responses.  It incorporates a novel method for recursive use of goroutines with a self-managing pool.  It is actively developed and supported by myself.  Thanks for your consideration in adding this to the awesome-go list.

Scott Johnston
johnston@vectaport.com

Documentation:  https://godoc.org/github.com/vectaport/flowgraph
Wiki: https://github.com/vectaport/flowgraph/wiki
Testing:  https://github.com/vectaport/flowgraph_test

### DIFF
--- a/README.md
+++ b/README.md
@@ -277,6 +277,7 @@ Please take a quick gander at the [contribution guidelines](https://github.com/a
 
 *Packages that help with building Distributed Systems.*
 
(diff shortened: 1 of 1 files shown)


shurcooL:
LGTM.

PR #569, opened 2015-09-08
Fix hash reference to ebooks section.

rajibahmed:


shurcooL:
The current link works correctly on github, the new one does not:

https://github.com/avelino/awesome-go#e-books

I'm guessing this was a meant to fix a problem on the website version? That should be done without breaking the github link.

rajibahmed:
it does not work on github too. Broken for both the website and github.

shurcooL:
It does work on GitHub for me. Its source confirms it:

```
<h2><a id="user-content-e-books" class="anchor" href="#e-books" aria-hidden="true"><span class="octicon octicon-link"></span></a>E-Books</h2>
```

PR #575, opened 2015-09-09
Remove 'doc', which is deprecated.

ndahlquist:
"Doc is an old program that was used to design a nice command-level API for presenting Go documentation. It is deprecated; use the new go doc in 1.5 instead."

- https://godoc.org/robpike.io/cmd/doc


matrixik:
Thank you for contribution. Removed.

PR #576, opened 2015-09-09
Added REX-Ray

clintonskitson:


avelino:
To create a new category needs to be more than 3 packages

PR #577, opened 2015-09-14
Update README.md

kpashka:
Added links to Slack / Telegram API


shurcooL:
LGTM.

Both packages look nice, implement known APIs, have docs, and aren't already present.

PR #579, opened 2015-09-18
Adds query language section.

chris-ramon:
I think we need a Query Language section to list `graphql` related libraries.

Adds: 
 - Query Language section.
 - `graphql-go` library within Query Language.


avelino:
To create a new category needs to be more than 3 packages

chris-ramon:
> To create a new category needs to be more than 3 packages

Ok, thanks for feedback.

I added a couple of active `graphql`related libraries.

PR #580, opened 2015-09-18
add more popular go projects in github

smallnest:


drewwells:
I can't speak for the others but gorilla mux should totes be in the list

GoranP:
True i agree

shurcooL:
Yeah, I was a little surprised it wasn't already there.

PR #581, opened 2015-09-20
Add GitHub Trending to Resource section

icaroseara:


shurcooL:
The link is good to add, but I don't think it should go in the header, uncategorized like that.

How about moving it to the Resources/Websites subcategory?

image

https://github.com/avelino/awesome-go#websites

shurcooL:
Actually, it's already there.

image

That's why the [Travis build](https://travis-ci.org/avelino/awesome-go/builds/81285190) failed with:

```
2015/09/20 19:47:10 duplicated link 'https://github.com/trending?l=go'
```

Closing because this link is already added.

shurcooL:
If you'd like to improve the wording of the existing entry so it's easier to spot, feel free to make another PR with that change.

PR #586, opened 2015-09-23
add command line interface: Readline

chzyer:


avelino:
```
2015/09/23 13:06:06 expected 'mitchellh/cli' but actual is 'readline'
2015/09/23 13:06:06 expected 'readline' but actual is 'mitchellh/cli'
```

chzyer:
@avelino, Hi~ Is any problems here?

PR #598, opened 2015-10-01
remove unmaintained package

norwack:
Remove `plubi` as I no longer maintain it and much better alternatives are available (I think).


shurcooL:
LGTM, this is your own package.

Thank you @norwack.

PR #612, opened 2015-10-08
Add mow.cli, a library for building CLIs

jawher:
A [library](https://github.com/jawher/mow.cli) for building CLI applications.

## Why yet another CLI library ?

mow.cli comes with support for what you would expect from any decent CLI library: posix syntax, option groups, commands and sub commands, contextual help, ...

What sets it apart is a sophisticated call validation, by providing a spec string to specify how exactly a command should be called.
For example, here's the spec string you can use with mow.cli to define the call syntax of a `cp` like command:

```
[-R [-H | -L | -P]] [-fi | -n] [-apvX] SRC... DST
```

I've written 2 articles to explain in more details the why and how: [1](https://jawher.me/2015/01/13/parsing-command-line-arguments-shameless-plug-mowcli/), [2](https://jawher.me/2015/01/18/parsing-command-line-arguments-finite-state-machine-backtracking/)

mow.cli is fairly well tested (84% coverage), well documented (see [godoc](https://godoc.org/github.com/jawher/mow.cli)) and the code quality is not too bad either ([A+ grade](http://goreportcard.com/report/jawher/mow.cli)).

### DIFF
--- a/README.md
+++ b/README.md
@@ -126,6 +126,7 @@ Please take a quick gander at the [contribution guidelines](https://github.com/a
 * [kingpin](https://github.com/alecthomas/kingpin) - A command line and flag parser supporting sub commands.
 * [liner](https://github.com/peterh/liner) - A Go readline-like library for command-line interfaces.
 * [mitchellh/cli](https://github.com/mitchellh/cli) - A Go library for implementing command-line interfaces.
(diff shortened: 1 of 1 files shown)


jawher:
Updated the description to add more details.

shurcooL:
LGTM.

PR #614, opened 2015-10-10
Add factory-go

bluele:
A fixtures replacement inspired by factory_boy and factory_girl.


avelino:
to create a new category must have at least 3 libs

PR #624, opened 2015-10-14
Delete Gaudi

icebourg:
Gaudi is being discontinued: http://marmelab.com/blog/2014/11/06/retiring-gaudi.html


shurcooL:
Thanks.

LGTM.

PR #635, opened 2015-10-20
Add Goclipse entry

bruno-medeiros:


shurcooL:
Can someone who uses Eclipse and this plug-in confirm it's worth adding?

I don't use Eclipse. But I've looked at the repo and it looks like a good plug-in at a glance.

There's nothing else wrong, so LGTM unless someone more familiar can add other feedback.

avelino:
Don't use eclipse, tested it and it worked perfectly.

PR #636, opened 2015-10-21
go-trigger added.

sadlil:
An event handling package added.


avelino:
must have at least 3 libs to create a category!

sadlil:
Oh. I didn't know that. Which category i can add this?

PR #647, opened 2015-10-24
Adding "glow", and a section for actual distributed systems rather than just libraries

chrislusf:


avelino:
What has been proposed in this change?

chrislusf:
For distributed systems, there were only libraries to build distributed systems. Here I am adding a distributed computing system, glow, which is similar to Hadoop, Spark, etc.

Any better suggestions?

PR #653, opened 2015-10-27
Added Slack RTM API

james-bowman:
Added Slack RTM (Real Time Messaging) API


avelino:
It would be interesting to your test project

PR #664, opened 2015-11-02
Add webhooks library for GitHub

joeybloggs:
* Add Github Webhook library, more service to come i.e. BitBucket..
 https://github.com/go-playground/webhooks

Most other libraries only return part of the webhooks even, this returns the WHOLE posted event data.


shurcooL:
I don't think this is the right section, since it deals with APIs, I think it should go where the go-github library is.

By the way, does go-github not provide this functionality?

joeybloggs:
Maybe, but I don't think "Libraries for accessing third party APIs." is quite right either as I'm not accessing any API but rather receiving events via an HTTP POST request.

go-github may have some of this functionality, I know they have the struct defined, however don't see them listening for webhook events. 

Regardless, the reason I created this was for ease of use, flexibility and to be able to easily add other services such as BitBucket, which I'm sure go-github is more strictly for GitHub.

joeybloggs:
I can move it there if you wish, what do you think?

joeybloggs:
OK moved it to the "Libraries for accessing third party APIs." section

shurcooL:
LGTM.

PR #668, opened 2015-11-10
Replace outdated Snapchat library with an updated one.

hako:
The previous Snapchat library is a year old and out of date. This library is up to date, progressing fast and is actively maintained.


README.md
```diff
@@ -770,7 +770,7 @@ See [go-hardware](https://github.com/rakyll/go-hardware) for a comprehensive lis
 * [shopify](https://github.com/rapito/go-shopify) - Go Library to make CRUD request to the Shopify API.
 * [slack](https://github.com/nlopes/slack) - Slack API in Go.
 * [smite](https://github.com/sergiotapia/smitego) - Go package to wraps access to the Smite game API.
-* [snapchat](https://github.com/jamieomatthews/gosnap) - Go wrapper for the snapchat API
+* [snapchat](https://github.com/neuegram/ghost) - Go Library for the Snapchat API.
```
shurcooL: I believe the name of the new package is `ghost`, not `snapchat`. So the title of this entry should be changed, and it'll have to be re-arranged so it's still alphabetical.

I agree that the proposed library looks better and can replace the previous one.

PR #677, opened 2015-11-19
Add goreturns to code analysis section

hanjianwei:


shurcooL:
The import path of this package is `sourcegraph.com/sqs/goreturns`, so I believe that would be a better link to use:

https://sourcegraph.com/sqs/goreturns

@sqs, can you confirm what the canonical URL to the home page of goreturns (not necessarily the same as its import path, but possibly so)?

sqs:
Correct!

shurcooL:
@hanjianwei, please update to the canonical home page URL and this can be merged afterwards.

hanjianwei:
@shurcooL @sqs 

I've updated the URL.

Sorry for my mistake!

shurcooL:
No worries, thanks for fixing it. LGTM.

PR #686, opened 2015-11-23
Update README.md

mistbow-git:
Add Ozzo framework.


avelino:
The word "legendary" is has a typo above.

It would be better to use more neutral, objective language, like "written by author of yii PHP framwork."

PR #706, opened 2015-12-07
Add gomega - an rspec-like assertion library

glestaris:


shurcooL:
LGTM.

This is not quite the same as ginkgo, right? Because ginkgo is already listed here.

glestaris:
Ginkgo is the test framework (structuring and running tests) while gomega
is the assertions/matchers library. They can be used independently of each
other.

shurcooL:
That's what I thought, just wanted to confirm. Thanks.

PR #707, opened 2015-12-09
Remove Martini.

shurcooL:
According to https://github.com/go-martini/martini/commit/15a47622d6a9b3e6a1eaca2681e4850f612471ea:

> The martini framework is no longer maintained.


dgryski:
Seems a bit premature; but OTOH nobody else has spoken up and said "I'll take it..."

shurcooL:
I'm okay with waiting on this PR, if desired. I just made it directly instead of adding it a future TODO list. It usually takes some time before PRs get reviewed too.

That said, it has been 25 days since that commit, so it's not like "it just happened".

PR #711, opened 2015-12-11
add drone - ci platform

patdhlk:


avelino:
@patdhlk of the 3 items you have in the list only the done.io which is a system, I prefer to keep "Tools for help with continuous integration"

patdhlk:
@avelino i deleted the section and moved drone.io back to 'Tools for help with continuous integration'

PR #721, opened 2015-12-20
Add Micro-service section

tiaguinho:


avelino:
to create a new category must have at least 3 libraries

PR #725, opened 2015-12-22
margelet and go_mediainfo added

zhulik:


avelino:
http://goreportcard.com/report/zhulik/go_mediainfo

lint very very down

zhulik:
Sorry, all issues was in tests, go fmt and golint doesn't check it by default. 

Fixed!

PR #734, opened 2015-12-26
Adds another SQL query builder called 'sqrl'

elgris:
This pull request adds one more SQL query builder which is actually a fork of famous [Squirrel](https://github.com/Masterminds/squirrel) with improved performance. Benchmarks included in the repo of the project ;)


avelino:
duplicate link

Issue #683, opened 2015-11-22
Automate the acceptance of high quality repos 


levigross:
Hi,

I prepose that you should consider automating the merging of certain entries. The hard part is using the proper metrics.

I would like to start the conversation by suggesting the following metrics. 
- Github stars
- Test coverage
- The output of http://goreportcard.com/ (which checks `go vet`, `go lint`, and `go fmt`)

joeybloggs:
Just my 2 cents,

* stars should not be taken into account for 2 reasons:
  * just because it has a bunch of stars does not mean it's any good.( just being realistic)
  * I've submitted pull requests and had packages added with very few stars if any, the whole point was to get some exposure, which stars may be a by-product

*  Test coverage for sure ,but what percentage, may have to be as low as 50-60% as sometimes some things cannot be faked reliably I.e. read failure from file on disk.

* Goreportcard is a great idea, the only thing that would concern me is the cyclomatic checking that site does, I don't know at what complexity to it starts showing issues, but I have a project where it shows issues where there are none.

levigross:
@joeybloggs 

I agree with your sentiments on Github stars. My opinion is based on the following two points:

1. Sometimes you have a product that is popular but doesn't have great test coverage: e.g Docker. 
1. Popular packages benefit from the community (and therefore over time and with the right management, their code quality gets really good)

As for test coverage, I am not suggesting that we only accept projects with 90% code coverage, but I do think it is important to assess a project based on its code coverage. The exact percentage should be determined based the list's existing coverage.

Additionally, I would like to suggest another metric we can use, longest unanswered Github issue. That will help us understand if the project is alive or not. 

Finally, I feel that it is important to remember, we want to *automate* the acceptance of items. We should validate whatever metrics we use by feeding the current list into the "automator" and seeing what comes out.

shurcooL:
I agree that this is an effort worth investing into, because it seems the current amount of inbound Pull Request is higher than the capacity of the contributors to accept it. There are many, many PRs that go unreviewed for weeks, if not months, which is a problem.

jgoodall:
Also throwing in my 2 cents... This site is described as "A curated list of awesome Go frameworks, libraries and software". _Curated_ implies that someone is actually reviewing the packages that get added to the list. If you start talking about automation, and there is no review, the _curated_ probably needs to be dropped. 

[...]

PR #752, opened 2016-01-07
adds `conform` to `Forms`

leebenson:
- adds [conform](https://github.com/leebenson/conform) to "Forms" section

### DIFF
--- a/README.md
+++ b/README.md
@@ -364,6 +364,7 @@ Please take a quick gander at the [contribution guidelines](https://github.com/a
 
 * [bind](https://github.com/robfig/bind)  - Bind form data to any Go values
 * [binding](https://github.com/mholt/binding) - Binds form and JSON data from net/http Request to struct.
+* [conform](github.com/leebenson/conform) - Keeps user input in check. Trims, sanitizes & scrubs data based on struct tags.
 * [formam](https://github.com/monoculum/formam) - decode form's values into a struct.
 * [forms](https://github.com/albrow/forms) - A framework-agnostic library for parsing and validating form/JSON data which supports multipart forms and files.
 * [gorilla/csrf](https://github.com/gorilla/csrf) - CSRF protection for Go web applications & services.
--- a/README.md
+++ b/README.md
@@ -364,7 +364,7 @@ Please take a quick gander at the [contribution guidelines](https://github.com/a
 
 * [bind](https://github.com/robfig/bind)  - Bind form data to any Go values
 * [binding](https://github.com/mholt/binding) - Binds form and JSON data from net/http Request to struct.
-* [conform](github.com/leebenson/conform) - Keeps user input in check. Trims, sanitizes & scrubs data based on struct tags.
+* [conform](https://github.com/leebenson/conform) - Keeps user input in check. Trims, sanitizes & scrubs data based on struct tags.
 * [formam](https://github.com/monoculum/formam) - decode form's values into a struct.
 * [forms](https://github.com/albrow/forms) - A framework-agnostic library for parsing and validating form/JSON data which supports multipart forms and files.
 * [gorilla/csrf](https://github.com/gorilla/csrf) - CSRF protection for Go web applications & services.


avelino:
review Gocyclo pls http://goreportcard.com/report/leebenson/conform

joeybloggs:
IMHO I don't think Gocyclo should be a factor in adding anyones library, although it **can sometimes** show a real issue, but allot of times it gives false positives when no issue really exist.

on a side note I don't even know what the threshold is set to when goreportcard.com runs gocyclo 5, 10, 20? it just seems very subjective to me.

avelino:
```
Gocyclo calculates cyclomatic complexities of functions in Go source code. The cyclomatic complexity of a function is calculated according to the following rules: 1 is the base complexity of a function +1 for each 'if', 'for', 'case', '&&' or '||'

/github.com/leebenson/conform/conform.go
Line 114: 19 conform Strings repos/src/github.com/leebenson/conform/conform.go:114:1
```

leebenson:
@avelino the library is iterating through reflected structs and parsing tag syntax. If you have a suggestion for reigning in loops and switches, I'd be happy to adjust the code but otherwise, the goals of gocyclo and the purpose of my code seem diametrically opposed. Artificially refactoring away from a single function would be solely to appease a third-party linter.

joeybloggs:
@avelino I understand the rules that count towards the complexity, but there is a threshold that the program has to hit before it starts thinking there are too many +1's denoted by the rules I think it's the over parameter here https://github.com/fzipp/gocyclo/blob/master/gocyclo.go#L40

but I don't see the number that is being used by goreportcard.com anywhere, besides I thinkit's up to a programmer to ultimately decide if the complexity is a problem or not.

leebenson:
forgive my bias, but I agree with @joeybloggs 

gocyclo is, at best, a somewhat arbitrary preemptor of code traps, based on a subjective score. My library is single purpose, designed to iterate through embedded structs, find qualifying strings, and format based on 16 possible user-specified rules. If gocyclo is choking on 19 counts, then that one switch statement accords for 84% of them. In some circumstances this might hint at the need to reduce complexity, but in my case, I see no more performant way for my code to function.

Since I do not know the N value that goreportcard considers "over", refactoring my code to satisfy an unknown quantity would be an exercise of appeasement to a subjective (and unknown) standard. It would be the same code, with a few extra function calls, likely rendering it less performant and increasing the surface area of an otherwise single-purpose function.

I can't see that upping the gocyclo score would increase the library's attractiveness to your user base. I appreciate your consideration for my lib's inclusion all the same.

leebenson:
I would also add that sqlx, one of the main SQL libraries in Go, only weighs in at a 60% gocycle score:
http://goreportcard.com/report/jmoiron/sqlx

Likely because of similar struct parsing.

joeybloggs:
so @avelino have you thought about Gocyclo?

avelino:
In this "PR" we're not talking about just Gocyclo
https://goreportcard.com/report/leebenson/conform

If you really want to get in that discucao should read more about "awesome-go"
https://github.com/avelino/awesome-go/issues/683

joeybloggs:
@avelino I never said you weren't just talking about Gocyclo for this pull request, I just asked if you have thought about it based on all of the previous comments in this thread about it.

I already brought it up in #683 in the first comment...

leebenson:
Actually when I first submitted, gocyclo was the only issue. The gofmt stuff is news to me and very trivial to fix- I probably had 'format on save' turned off in Atom on my last commit.

I'm not sure what you have against this package or why you care so much about what a mindless linter (that's not designed for the use case the package intends to solve) determines the code quality to be?

If you don't feel the package is useful to readers, that's a different story.

avelino:
Gentlemen, we get a lot of PR, if you don't follow the minimum quality (now use goreportcard to help us) does make it hard to maintain this list.

You want to help me send the quality of list? We need many employees.

leebenson:
I mean 'mindless' in the sense of 'not context aware', to be clear. It's a decent first pass, but when a package's entire purpose defeats the lining, lining is a secondary concern.

There are plenty of packa

leebenson:
Thank you @avelino

joeybloggs:
Sure thanks @avelino

PR #776, opened 2016-01-22
Add gogs

yejijie:
Gogs (Go Git Service) is a painless self-hosted Git service.


shurcooL:
The link you've included to gogs is a binary or a project, it is not a library (although it contains many libraries).

It is Go software, so it can fit under https://github.com/avelino/awesome-go#tools section.

yejijie:
@shurcooL ok

shurcooL:
gogs has already been added long ago in #24, and it's listed under https://github.com/avelino/awesome-go#devops-tools. Closing.

PR #785, opened 2016-01-30
add ToTo

blogcin:
ToTo is a simple proxy server written in Go language


matrixik:
Thank you for contribution.

Non alphabetical. No license?
http://goreportcard.com/report/blogcin/ToTo
What is the purpose? Examples and more info in README would be great.

PR #804, opened 2016-02-04
Add Mergo

imdario:


matrixik:
Thank you for contributing.
Can you fix some small warnings before merging:
https://goreportcard.com/report/imdario/mergo

imdario:
All fixed but gocyclo warnings will take longer. I'll check against upstream [Go repository](https://github.com/golang/go/blob/master/src/reflect/deepequal.go) and to make sure any improvement is merged in Mergo's implementation.

If the current score is fine, consider merging this PR. Thanks!

matrixik:
Thank you for all fixes and for working on this useful package.
Some people don't like `gocyclo` idea but personally I think it make code easier to comprehend.

PR #811, opened 2016-02-10
Added grab

cavaliercoder:
Go package for managing file downloads


matrixik:
Thank you for contribution.
Would be nice if, with time, you can increase test coverage: https://gocover.io/github.com/cavaliercoder/grab

cavaliercoder:
Thanks for adding grab!
I haven't seen gocover.io before. I'll be making good use of it, thanks very much!

matrixik:
You can also find [gometalinter](https://github.com/alecthomas/gometalinter) helpful.

PR #821, opened 2016-02-16
Adding a link to Moldova, a utility for generating random data based …

StabbyCutyou:
…on an input template

I placed it into Utilities, but it could also go under either Templates or Testing. The primary use is as a binary, but it's also capable of being used as a library directly.


matrixik:
Looks nice, thank you for contribution.

PR #829, opened 2016-02-21
Add go-twitter to third-party APIs

dghubble:
* https://github.com/dghubble/go-twitter
* https://dev.twitter.com/overview/api/twitter-libraries#go

### DIFF
--- a/README.md
+++ b/README.md
@@ -846,6 +846,7 @@ See [go-hardware](https://github.com/rakyll/go-hardware) for a comprehensive lis
 * [ghost](https://github.com/neuegram/ghost) - Go Library for accessing the Snapchat API.
 * [github](https://github.com/google/go-github) - Go library for accessing the GitHub API.
 * [go-marathon](https://github.com/gambol99/go-marathon) - A Go library for interacting with Mesosphere's Marathon PAAS
+* [go-twitter](https://github.com/dghubble/go-twitter) - Go client library for the Twitter v1.1 APIs.
 * [goamz](https://github.com/mitchellh/goamz) - Popular fork of [goamz](https://launchpad.net/goamz) which adds some missing API calls to certain packages.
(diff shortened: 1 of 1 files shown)


shurcooL:
Do you know how it compares with https://github.com/ChimeraCoder/anaconda?

dghubble:
I'm the author who wrote `go-twitter` while a Twitter employee, so I'll try not to create debate here.

Both provide access to a similar set of APIs. . `go-twitter` tends to favor composability at the expense of not providing as many bells and whistles out of the box. For example, it doesn't choose an OAuth1 dependency, takes an `http.Client` directly, and supports multiple users as a result. It is designed much like `google/go-github`. `anaconda` tends to make choices for users, which can be convenient (handles request throttling for a user).` Same APIs, different designs and priorities (imo).

Both are listed in the Twitter Go Clients list https://dev.twitter.com/overview/api/twitter-libraries#go

dghubble:
Many links in the list are named by the third party service. Here, it might be reasonable to label this `go-twitter` instead.

shurcooL:
> Many links in the list are named by the third party service. Here, it might be reasonable to label this `go-twitter` instead.

I think that's a mistake, they should be named after the library name. So yeah, it should be `go-twitter` (which will need to be placed elsewhere due to alphabetization).

> It is designed much like `google/go-github`.

Awesome, thank you for describing the differences! That sounds awesome, I really like `google/go-github` style and I'm glad to hear `go-twitter` follows it (similarly to [`go-travis`](https://github.com/Ableton/go-travis#disclaimer)).

shurcooL:
If you rename it to from `twitter` to `go-twitter` (to match the library name) and ensure correct alphabetize order, this can be merged.

shurcooL:
LGTM. If it's ready from your side, I'll merge?

dghubble:
:+1:

shurcooL:
Thanks for contributing, this looks like an awesome Go package!

PR #835, opened 2016-02-23
Add llb

kirillDanshin:
llb is a low-load-backend - it's created for fast redirects and 404 errors, that can be detected on proxy level. I'm using it with traefik - all requests that are not matched any traefik's frontends are passed to llb redirect so user redirected to another site very fast.
It's not an advanced backend but it's useful IRL.


matrixik:
Thank you.

PR #841, opened 2016-02-25
Add rqlite, replicated SQLite using Raft consensus

otoolep:


matrixik:

PR #843, opened 2016-02-25
feat: add gentleman package

h2non:
Repo: https://github.com/h2non/gentleman


kirillDanshin:
hey, it's looks awesome! when we can wait for v1?

h2non:
I would like to have it under v0 for a while before moving to v1. However now the API is stable should not break any interface.

matrixik:

PR #826, opened 2016-02-19
Added Golf to Go web frameworks.

dinever:
Hi,

I've just created another Go web framework named Golf.

Thanks

PR #850, opened 2016-03-02
Added Golf to Go web frameworks.

dinever:
Hi,

I've created another Go web framework named Golf.

We have discussed it at #826 , and I have improved it after that:

https://goreportcard.com/report/github.com/dinever/golf

I am still working on the test coverage, so it would be great if you can add my project.

Thanks


matrixik:
Thank you for contribution, looks better now.

PR #852, opened 2016-03-02
Add github.com/rafael-santiago/cherry

rafael-santiago:
Just adding a new messaging software to the previous collection.


matrixik:
Thank you for contribution.
First: `Messaging` section is for libraries, you created main program so I think [Other Software](https://github.com/avelino/awesome-go#other-software) is more appropriate section for it.
Second: please don't use relative import paths, they break `go get` and many other useful tools (check here for more info: https://www.reddit.com/r/golang/comments/2588lk/relative_imports_do_or_dont/ )

Other than that looks nice.

rafael-santiago:
I have done the suggested improvements.

The link was moved to the "Other Software" section and now the application's sources are using the GOPATH instead of the relative paths for the imports.

Thank you for your suggestions and tips.

matrixik:
Thank you.

PR #854, opened 2016-03-03
Add gock package

h2non:
gock - https://github.com/h2non/gock


matrixik:
Nice, thank you for contributing.

PR #857, opened 2016-03-07
add alxrm/ugo

alxrm:
Link to [goreportcard](https://goreportcard.com/): https://goreportcard.com/report/github.com/alxrm/ugo

Link to [gocover](https://gocover.io/): https://gocover.io/github.com/alxrm/ugo

Link to [godoc](https://godoc.org/): https://godoc.org/github.com/alxrm/ugo

ugo is a set of slice helper methods, written in Go. It provides many different functions, mostly for slices though. Inspired by [underscore.js](underscorejs.org) and copies it's syntax and behaviour.

### DIFF
--- a/README.md
+++ b/README.md
@@ -952,6 +952,7 @@ See [go-hardware](https://github.com/rakyll/go-hardware) for a comprehensive lis
 * [sling](https://github.com/dghubble/sling) - Go HTTP requests builder for API clients.
 * [spinner](https://github.com/briandowns/spinner) - Go package to easily provide a terminal spinner with options.
 * [sqlx](https://github.com/jmoiron/sqlx) - provides a set of extensions on top of the excellent built-in database/sql package
+* [ugo](https://github.com/alxrm/ugo) - ugo is slice toolbox with concise syntax for Go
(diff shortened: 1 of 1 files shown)


matrixik:
Thank you for contribution.

`golint` score is too low. Docs looks OK (thought they could easily conform to `golint`), but other warnings should be addressed before merging.

alxrm:
@matrixik Thanks for your help! I refactored my code and docs for them to conform `golint` check

matrixik:
Thank you for fixes.

Some last comments:
I leave comment in https://github.com/alxrm/ugo/commit/441abe0e640ef545c5127960fe49dbb89601fcc8#commitcomment-16590575

And you don't need this line: https://github.com/alxrm/ugo/blob/master/ugo_test.go#L26
In tests you have access to all methods (also private) by default.

Do you need to store **.idea** folder in your repo?

alxrm:
@matrixik I cannot access those methods without import, because tests are in the different package (ugo_test). I saw this in builtin packages like [Sort](https://github.com/golang/go/tree/master/src/sort)

No I don't, I will add **.idea** folder to .gitignore

I appreciate your reviews, they are really helpful, thank you!

matrixik:
Oh, right, you are using black-box testing here. My fault.

Thank you.

PR #860, opened 2016-03-10
Adding xo to Database section of README.md

kenshaw:
Adding the xo command line database schema/query code generation tool to Database section of README.md.


kenshaw:
If there are issues with this, I would appreciate the feedback. Thanks.

PR #889, opened 2016-04-03
Adding Tile38

tidwall:
[Tile38](https://github.com/tidwall/tile38) is a geolocation data store, spatial index, and realtime geofence. It supports a variety of object types including lat/lon points, bounding boxes, XYZ tiles, Geohashes, and GeoJSON.

It's like Redis but for GIS.

- website: http://tile38.com
- godoc.org: https://godoc.org/github.com/tidwall/tile38
- goreportcard.com: https://goreportcard.com/report/github.com/tidwall/tile38

### DIFF
--- a/README.md
+++ b/README.md
@@ -234,6 +234,7 @@ Please take a quick gander at the [contribution guidelines](https://github.com/a
 * [rqlite](https://github.com/otoolep/rqlite) - Replicated SQLite, using Raft consensus.
 * [tidb](https://github.com/pingcap/tidb) - TiDB is a distributed SQL database. Inspired by the design of Google F1.
 * [tiedot](https://github.com/HouzuoGuo/tiedot) - Your NoSQL database powered by Golang.
+* [Tile38](https://github.com/tidwall/tile38) - A geolocation DB with spatial index and realtime geofencing.
 
 *Database tools.*
 


kirillDanshin:
You forgot to add all gocover.io links.
I'll do it for you as an exception:

[tidwall/tile38/client](https://gocover.io/github.com/tidwall/tile38/client)
[tidwall/tile38/cmd/tile38-cli](https://gocover.io/github.com/tidwall/tile38/cmd/tile38-cli)
[tidwall/tile38/cmd/tile38-server](https://gocover.io/github.com/tidwall/tile38/cmd/tile38-server)
[tidwall/tile38/controller](https://gocover.io/github.com/tidwall/tile38/controller)
[tidwall/tile38/controller/bing](https://gocover.io/github.com/tidwall/tile38/controller/bing)
[tidwall/tile38/controller/collection](https://gocover.io/github.com/tidwall/tile38/controller/collection)
[tidwall/tile38/controller/log](https://gocover.io/github.com/tidwall/tile38/controller/log)
[tidwall/tile38/controller/server](https://gocover.io/github.com/tidwall/tile38/controller/server)
[tidwall/tile38/core](https://gocover.io/github.com/tidwall/tile38/core)
[tidwall/tile38/geojson](https://gocover.io/github.com/tidwall/tile38/geojson)
[tidwall/tile38/geojson/geo](https://gocover.io/github.com/tidwall/tile38/geojson/geo)
[tidwall/tile38/geojson/geohash](https://gocover.io/github.com/tidwall/tile38/geojson/geohash)
[tidwall/tile38/geojson/poly](https://gocover.io/github.com/tidwall/tile38/geojson/poly)
[tidwall/tile38/index](https://gocover.io/github.com/tidwall/tile38/index)
[tidwall/tile38/index/qtree](https://gocover.io/github.com/tidwall/tile38/index/qtree)
[tidwall/tile38/index/rtree](https://gocover.io/github.com/tidwall/tile38/index/rtree)

@matrixik There are some vendor packages, I think that don't need to be covered by maintainer.

kirillDanshin:
@tidwall thank you for contribution. 
I'll merge it, but please fix low test coverage as soon as possible.
Also, can I do realtime tracking of many (~2-4 millions) clients with ability to detect when any client is close to some other with your project?

tidwall:
@kirillDanshin Huge thanks for including my project!! I'll work on improving coverage.

Yes, the geofencing feature can track millions of changing records and detect and notify you in realtime when those records enter or exit a boundary, radius, or polygon.

kirillDanshin:
@tidwall 

in this case I want to do this:
lets say that we have 3 millions people, including Alice and Bob.
we want to say Alice that Bob is close (lets say, 300 meters from her) and we may want to meet.
we have 3 millions people that may want to meet each other when they close.
can I get this somehow with this package?

tidwall:
Very interesting. Kind of like roaming geofences. I think this might work.

As it stand right now, you could create a point named Alice and also hooked geofence named Alice300.

```
SET persons alice POINT 34 -118
SETHOOK alice300 https://10.0.1.15/endpoint NEARBY persons MATCH bob DETECT inside FENCE POINT 34 -118 300
```

When the point named "bob" from collection "persons" get within 300 meters from Alice, a message gets fired off to "https://10.0.1.15/endpoint". 

Then each time Alice's location changes, both the point and hooked fence would need to be updated.

---

While this solution is technically feasible I'm concerned about the server load, because Geofences where intended to act as lightweight network connectors for **static** locations, not moving. Millions of SETHOOK calls will cause a ton of memory chatter and a ton of extra cpu cycles.

Ideally we would send a single ```SETHOOK``` that is bound to a point by it's name and a defined proximity. Then all ```SET ... POINT``` calls will automatically update the hook. 

---

Sorry for blabbering on, but this is such an interesting feature. I plan on looking into more.

Thanks a bunch for your suggestion.

kirillDanshin:
@tidwall yes, but it's not covering *all* that I want to say. I think, I need to repeat that `we have 3 millions people that may want to meet each other when they close`. So each user is our Alice and each other user is our Bob. For example we want to make users' meetings simpler even if we wasn't knew that they can be close to each other.
Now your software can help if we want to send user a message when he, for example, in a mile from our shop and say him "Hi! Come to us in a hour and get 25% off!".
I want to send a message when user A close to user B without need to know who is user A and who is user B.
---
I think ideally we would send a single *you_name_event* when `point_a` in a defined proximity to `point_b` with `point_a_name`, `point_a_coords`, `point_b_name`, `point_b_coords`, `proximity` and `speed` in the event data.

I just want to recieve in some endpoint data about points from collecion when they gets closer then defined proimity.

tidwall:
@kirillDanshin I see. I was over thinking it. I don't believe the geofencing is needed at all. 

I suggest simply sending a SET and NEARBY commands for every time a users location changes.

```
SET users userA POINT 34 -118
NEARBY users POINT 34 -118 300
```

This will update the userA point and then return back all users within 300 meters of that point, regardless who those users are. 

This is an fast operation for Tile38. I just tested by adding 3,000,000 random users around the Silicon Valley area. I then updated and searched around a single user. It resulted in a list of 74 users and took 250 microseconds of server time. It took about 40 secs to insert 3 million random users. The server uses around 1.1 GB  of ram to store that many points.

[Here's a Gist of the commands I used](https://gist.github.com/tidwall/0c0fe655ef9e15c9f6b9918a3d70458b)

Would this work for your application?

kirillDanshin:
@tidwall nice gist, thanks. And I see that Tile38 is really fast. unhelpfully it will not work for me, because I'll need to search manually what users `nearby`. I'm looking for ability to subscribe once to all events like this. so I can create user with name that equals ID from my user database, for example, and it'll be really simple to notify all users that within 300 meters from each other.
I think that we can use geofencing to detect user's moves, search users that close to him on every user's move and trigger special hook if something changed.

tidwall:
In case you are still interested I just push roaming geofences functionality to 8fdc35a in the master branch. There's some details about the syntax at tidwall/tile38#17.

PR #918, opened 2016-04-18
Attach repo info for each GitHub repo

chaconnewu:
I created a separate file ("awesome_go_with_repo_info.md") that attach repo info (stars count, and last updated date) for each GitHub repositories included in README.md. It will help users quickly get the basic info of a certain repo.

I can update the file regularly (e.g. daily, weekly) or as needed.

View the file directly at:
https://github.com/chaconnewu/awesome-go/blob/attach_repo_info/awesome_go_with_repo_info.md

### DIFF
--- a/awesome_go_with_repo_info.md
+++ b/awesome_go_with_repo_info.md
@@ -86,37 +86,37 @@ Audio/Music
 
 *Libraries for manipulating audio.*
 
--   [flac](https://github.com/eaburns/flac) <span> | ★ 48, pushed 38 days ago | </span> - A native Go FLAC decoder.
--   [flac](https://github.com/mewkiz/flac) <span> | ★ 23, pushed 15 days ago | </span> - A native Go FLAC decoder.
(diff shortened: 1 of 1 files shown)


dixonwille:
You should be able to script this with github's api? That would be better than manually adding them. Possibly (just an idea and not sure if Travis-Ci can do this) but when a PR is made it runs the test. Maybe have the test gather the information and push to the PR branch the new changes for all repo's. Stars change and last commit changes a lot.

PS. I am not affiliated with this group at all. Just figured I can put the idea out there.

chaconnewu:
Thanks for the comment @dixonwille ! Yes I have scripts to do this, which run every day to gather the latest info. So the info attached is at most 24 hours outdated.

kirillDanshin:
I think we need to do it on a website. It is a bad idea to regenerate and merge it weekly.

any objections?

chaconnewu:
I wouldn't mind doing it on a website. But I'm wondering what do you think are the downsides for generating and merged it regularly? Thanks!

cassiobotaro:
Already discussed on #1213

PR #933, opened 2016-05-02
Added d3d9 library.

gonutz:
I created a Direct3D9 wrapper for Go because none existed. The API is complete, I created some samples, added a readme with build instructions and documentation, ran go vet and golint on the code and fixed the issues that needed to be fixed, tested the library on two different Windows systems with different compilers and now it is ready to be used by other people so I would like it to be added to this list so people can find it.
Here are the links:
[d3d9 library](https://github.com/gonutz/d3d9)
[godoc](https://godoc.org/github.com/gonutz/d3d9)
[goreportcard](https://goreportcard.com/report/github.com/gonutz/d3d9)
gocover is not able to build the project, I assume because it uses cgo on Windows.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1301,4 +1301,5 @@ Where to discover new Go libraries.
 
 ## Windows
 
+* [d3d9](https://github.com/gonutz/d3d9) - Go bindings for Direct3D9
 * [go-ole](https://github.com/go-ole/go-ole) - Win32 OLE implementation for golang.


kirillDanshin:
@gonutz hi. please, set up any other coverage service.

gonutz:
@kirillDanshin hi back :)
There are no tests in this library, this is a thin wrapper around the C API and thus not suited for unit tests. There are however samples that use the API and I am using it in other (private) projects to assure that it works.

kirillDanshin:
@gonutz hi.
how anybody can test that your lib work in target environment? I don't know. if you can tell me how and add it to your docs, I can review it. if not, I don't see why this lib needs to be listed in `awesome-go`.

gonutz:
@kirillDanshin
There is a 'Usage' part in the _readme_, there is a _samples_ folder containing five simple samples with build scripts and instructions, the _readme_ mentions my entry to the GopherGala hackathon - [Gophette](https://github.com/gonutz/gophette), a more involved example program which uses the library when building on Windows.
These are some ways to run the library in the target environment, which of course is Windows only.
If you are a curator of this project, please put some effort into it, it is insulting to me that you did not even seem to have read the _readme_ of my project before answering.
If on the other hand my docs are just not understandable then please point out to me the parts that need to be reviewed and I will fix the problems.

kirillDanshin:
@gonutz 
I've read your readme before answering. I've seen Gophette then.
Yes, your docs are just not understandable for me. And I still don't understand how I can check that the library are working as you expected, and how I can automate tests for my projects - who knows, maybe I found a bug in the library, but I'll think that I write something wrong because I can't check that the library is working as expected.
Now you understand why I want to know how I can check the whole lib?

gonutz:
@kirillDanshin
OK, now I see your problem, sorry for the harsh tone in my last reply then.

First, please point out to me the specific parts of the docs that you do not understand, so I can clarify them. English is not my first language so there might be some confusion.

To address your second concern, let me rephrase my explanation of why there cannnot be automated tests for this package: similarly to [go-gl](https://github.com/go-gl/gl) and [glfw](https://github.com/go-gl/glfw) this library is just a minimal wrapper around a C API. You do not see unit tests in those packages either and they are still listed.
The point is that these packages do not provide new functionality of their own in the Go code, they just forward all Go calls directly to C. If you look at my code you will see that all it does is take Go parameters, map them to C and then call into Direct3D9 via C.
What you would have to test is that the C code works, that is where the functionality is. This job is luckily done by Microsoft.
The only thing to test is that the Go function actually calls the right C function, this is of course the case. As stated in the _readme_ I created the wrapper with a little generator tool using Microsoft's online documentation.

There currently is no other library that lets you program to the Direct3D API from Go other then this one, which is why it should be in this list. There are multiple wrappers for OpenGL, but only targetting OpenGL for a desktop application will limit your potential user base to those who have installed the latest graphics drivers on their Windows machines. Windows only comes with OpenGL 1.1 support out of the box but has full Direct3D9 support in all versions from XP and above. This means having a Direct3D9 rendering path in your code largely increases the number of people able to run your application.

This is why I believe this library should be in the list. It is my contribution to make Go better suited for creating desktop applications on Windows.

kirillDanshin:
@gonutz I see. I've reviewed your package's source, so I'm sorry for taking so long time to reply. It took about six hours for me to review the code and prepare an answer in this case.
Some issues in the docs that I see:
- no comments for constants. they can be generated or just copied from d3d9 documentation.
- no examples in docs (I mean in [godoc](https://godoc.org/github.com/gonutz/d3d9))

Godoc comments are useful for people and in IDE that supports godoc for external packages.

gonutz:
@kirillDanshin
thank you for taking the time, you are right, the constants and the structs do not have documentation comments yet, I will fix this, the function comments are already taken from MSDN so I will just do the same for the rest.
What do you mean with examples in the docs?
Do you mean full running samples, the API is quite involved and it takes plenty of code to get something on the screen, so this might be overkill for the docs, I could however refer readers to the _samples_ folder of the project?

kirillDanshin:
@gonutz 
Yes, refer to the samples will be enough, I think.

gonutz:
@kirillDanshin
Alright, I will reply back when the changes are implemented.

gonutz:
@kirillDanshin
Hey, I have now commented the constants, enumerations and structs that had no comments. I copied them from Microsoft's online documentation semi-automatically.
I also refer users to the _samples_ folder now.
Thank you for these hints, this will indeed be useful to users, I agree.

kirillDanshin:
@gonutz thanks for your contribution. please, check out `go_vet` errors as soon as possible.
but for now, the package looks good to me. merged.

gonutz:
@kirillDanshin
thanks, the `go vet` warnings are actually not errors, those are the additional helper functions that I added so the user does not have to pass unsafe.Pointers when sending data to the graphics card. The C API uses `void*` pointers for that, so I created some helper functions for the usual cases, sending a slice of byte, float32 etc.
These functions are actually OK.

PR #948, opened 2016-05-11
add dog-tunnel

vzex:
add dog-tunnel, a p2p tunnel


kirillDanshin:
@vzex please create a new PR without the diff hell. we don't accept PR like this.
also, don't ignore our PR template. thanks.

PR #982, opened 2016-06-08
Update location/name of codegangsta/cli

meatballhat:


shurcooL:
Confirmed. See https://github.com/codegangsta/cli/pull/422/files for proof of org rename. LGTM.

Thanks!

PR #994, opened 2016-06-17
add gron package -- Cron Jobs in Go.

roylee0704:
Package links:

github.com:

https://github.com/roylee0704/gron

godoc.org:

https://godoc.org/github.com/roylee0704/gron

goreportcard.com:

https://goreportcard.com/report/github.com/roylee0704/gron

gocover.io:

https://gocover.io/github.com/roylee0704/gron

### DIFF
--- a/README.md
+++ b/README.md
@@ -972,6 +972,7 @@ See [go-hardware](https://github.com/rakyll/go-hardware) for a comprehensive lis
 * [gorequest](https://github.com/parnurzeal/gorequest) - Simplified HTTP client with rich features for Go.
 * [gotenv](https://github.com/subosito/gotenv) - Load environment variables from `.env` or any `io.Reader` in Go
 * [grequests](https://github.com/levigross/grequests) - An elegant and simple `net/http` wrapper that follows Python's requests library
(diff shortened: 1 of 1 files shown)


kirillDanshin:
hi.
[please](https://github.com/roylee0704/gron/blob/master/schedule_test.go#L191) never [don't](https://github.com/roylee0704/gron/blob/master/schedule.go#L52) [panic](https://github.com/roylee0704/gron/blob/master/schedule.go#L59).
keep calm and I'll merge it.
thanks.

roylee0704:
I assume that everything's okay? :)

roylee0704:
Thanks for the message, if you look carefully,

1 [don't](https://github.com/roylee0704/gron/blob/master/schedule.go#L52). It was intentionally placed (at the early stage) to assert that the preconditions of that function hold.

2. [panic](https://github.com/roylee0704/gron/blob/master/schedule.go#L52). Checking this error is unnecessary and burdensome if the caller knows that a particular call cannot fail, a good example would be `regexp.Compile`.

3. [please](https://github.com/roylee0704/gron/blob/master/schedule_test.go#L191). This is an helper function in test suite, there's nothing much to do if caller(test function) inputs an invalid date format.

PR #997, opened 2016-06-21
update go-playground/form's description to include Encoder functionality

joeybloggs:
Update [go-playground/form's](https://github.com/go-playground/form) description as it now also supports Encoding url.Values from a struct along with Decoding.


shurcooL:
LGTM, thank you.

PR #1003, opened 2016-06-23
Update README.md

hydroflame:
literally just removing my libs


shurcooL:
What's the reason you're removing them?

They are indeed your repos (and the links are all 404 now), so LGTM. Thank you!

PR #1018, opened 2016-07-09
Remove aac/h264 package because of 404.

### LINKED ISSUE #1016: Video - aac/h264 404
[aac/h264](https://github.com/go-av/codec)

shurcooL:
Resolves #1016.

As I mentioned there, @nareix, please let us know if the package has moved and is still deserving to be listed here, then we can update the link and close this PR.

According to https://github.com/avelino/awesome-go/blob/master/CONTRIBUTING.md#reporting-issues, we should wait 1 week.

I'm not sure if that applies if the entire repo is missing and the link is 404. Any thoughts?

### DIFF
--- a/README.md
+++ b/README.md
@@ -1035,7 +1035,6 @@ See [go-hardware](https://github.com/rakyll/go-hardware) for a comprehensive lis
 
 *Libraries for manipulating video.*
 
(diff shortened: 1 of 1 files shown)


shurcooL:
@avelino Did you see my comment about the CONTRIBUTING.md saying to wait 1 week? Any thoughts on if it's applicable here and why/why not?

PR #1060, opened 2016-08-16
Add Gitfile

bradurani:
Add Gitfile, a lightweight package manager for Git repos


bradurani:
Do you need something from me?

shurcooL:
The section you're adding to is called *Libraries for package and dependency management.*, but your project is a command, not a library.

https://godoc.org/github.com/bradurani/Gitfile/gitfile

bradurani:
Where should it go? It seems the command line section is also for libraries. This is a command line package manager

shurcooL:
There's a section for software and tools written in Go here:

https://github.com/avelino/awesome-go#tools

But you're right, this Package Management section contains package management tools, not necessarily libraries.

So your change is fine. We should update the description, because "Libraries for package and dependency management" is not accurate.

avelino:

PR #1079, opened 2016-09-01
Update Readme

SlevinT:
This is not my own package. A list of packages I found that quite often rely on others in the go-openapi accout.


kirillDanshin:
@SlevinT please fix the order. you need to pass `go test` before you merge. also you can use commit amend to minimize commits in your PR. thanks.

PR #1148, opened 2016-10-24
Replace Boom with Hey

jorinvo:
[Boom](https://github.com/rakyll/boom) has been renamed to [Hey](https://github.com/rakyll/hey) because of a name conflict with another binary.
The old URL is still working but users are encouraged to change to the new one.

Not sure if it's better to keep both, Boom and Hey, in the list for reference..


kirillDanshin:
Thanks @jorinvo.

jorinvo:
Keep up the good work on this awesome list :)

PR #1187, opened 2016-11-21
Removed clevergo

ruishengyang:
That package is no longer maintained, and will be delete in a few weeks.

Thanks you :)


kirillDanshin:
don't remove the repo! it will break others' code!
we can remove the link from awesome-go, but you should leave repo in `go get`able and buildable state.

ruishengyang:
Ok, I forked it to a new [account](https://github.com/headwindfly/clevergo), so it won't break others' code.

but i hope remove it from this list, thank you for reminding me.

kirillDanshin:
@ruishengyang if you'll remove the account you break code anyway because import paths depends on full repo url including username

ruishengyang:
@kirillDanshin You are right, so I created a new account(using the old username) and forked the repo(https://github.com/headwindfly/clevergo).

kirillDanshin:
@ruishengyang thanks.
merged :(

PR #1205, opened 2016-12-04
Fix anchor for generation-generic menu item

m0sth8:


kirillDanshin:
it works as expected and doesn't work with your changes.

m0sth8:
@kirillDanshin sorry, I didn't check github page. But is it expected to not work at all on the page http://awesome-go.com/#generation--generics ?

kirillDanshin:
@m0sth8 that what we should fix. but we should leave github page working.

Issue #1217, opened 2016-12-13
Update tests to show correct order.


zet4:
So recently I copied avelino/awesome-go structure for my own community listing for a service.
While dealing with helping people add new items because of the forced alpha sort I came up with a quick fix for it which can be seen in my [repo_test.go](https://github.com/beautiful-discord-community/resources/blob/master/repo_test.go),  line [95](https://github.com/beautiful-discord-community/resources/blob/master/repo_test.go#L95) in particular, basically it logs the correct order if there's an error, I also ported my test to 1.7 (I use t.Run in mine).

shurcooL:
That sounds like a useful thing. Are you interested in making a PR for it?

zet4:
If you don't mind me updating travis to 1.7 and rewriting it to use testing instead of log then sure

shurcooL:
I don't see why not. Go 1.7 is the current stable release of Go.

zet4:
I think you can see that in my PR but yes you do use log in your current testing.

PR #1219, opened 2016-12-13
Update test to 1.7 and display correct sorting order

zet4:
Per discussion in #1217 updating the test to latest stable and adding a log for correct sorting order if a fail occurs.

### DIFF
--- a/repo_test.go
+++ b/repo_test.go
@@ -25,16 +25,18 @@ func TestDuplicatedLinks(t *testing.T) {
 	links := make(map[string]bool, 0)
 
 	query.Find("body  a").Each(func(_ int, s *goquery.Selection) {
-		href, ok := s.Attr("href")
-		if !ok {
-			t.Errorf("expected '%s' to have href", s.Text())
-		}
+		t.Run(fmt.Sprintf("%s", s.Text()), func(t *testing.T) {
+			href, ok := s.Attr("href")
+			if !ok {
+				t.Error("expected to have href")
+			}
 
-		if links[href] {
-			t.Fatalf("duplicated link '%s'", href)
-		}
+			if links[href] {
+				t.Fatalf("duplicated link '%s'", href)
+			}
 
-		links[href] = true
+			links[href] = true
+		})
 	})
 }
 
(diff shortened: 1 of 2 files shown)


repo_test.go
```diff
@@ -27,14 +27,11 @@ func TestDuplicatedLinks(t *testing.T) {
...
-			log.Printf("expected '%s' href", s)
-			t.Fail()
+			t.Errorf("expected '%s' to have href", s.Text())
 		}
 
 		if links[href] {
-			log.Printf("duplicated link '%s'", href)
-			t.Fail()
-			return
+			t.Fatalf("duplicated link '%s'", href)
```
shurcooL: There was a `return` here before, but it's removed now. Is that intentional?

repo_test.go
```diff
@@ -46,7 +43,12 @@ func testList(t *testing.T, list *goquery.Selection) {
 		testList(t, items)
 		items.RemoveFiltered("ul")
 	})
-	checkAlphabeticOrder(t, list)
+
+	cat := list.Prev().Text()
```
shurcooL: What does `cat` mean?

zet4: `cat` contains name of the category that is to be sort checked

shurcooL: I see. How about naming it `category` then? That would be easier to understand.

zet4: yeah sure that sounds good, I pushed a commit to change that

repo_test.go
zet4: Regarding this exit I am not quite sure if it would terminate the entire test (as its running inside a `Each` function but probably still the same gorutine.

zet4: So I committed a fix for it

zet4:
@shurcooL any particular reason there hasn't been any updates on your side?

shurcooL:
No particular reason, I was sleeping at night, and I woke up recently. :) I plan to review this soon.

shurcooL:
If this looks good to another collaborator, let's merge this and test it out afterwards.

kirillDanshin:
LGTM, I'll merge this, let's test it out

PR #1278, opened 2017-01-31
Add OctoLinker browser extension

stefanbuck:
[github.com/OctoLinker/browser-extension](https://github.com/OctoLinker/browser-extension)

Navigate through projects on GitHub.com efficiently with the OctoLinker browser extension.

Most projects consist of many files and third party dependencies. Files are referencing other files and / or dependencies by language specific statements like `import` or `require`. OctoLinker makes these references clickable. No more copy and search.

### Core library imports

A standard library import like `import "math/big"` will reference to `https://golang.org/pkg/math/big`.

### Relative imports

An import path beginning with `./` or `../` is treated as a relative path. OctoLinker tries to resolve the path and redirects you to the linked resource.

### Remote imports

Last but not least, OctoLinker also supports hosting sites. For example `import "github.com/foo/bar"` will reference to `https://github.com/foo/bar`. The following hosting sites are supported:

- github.com
- bitbucket.org
- launchpad.net
- hub.jazz.net
- gopkg.in

**Note, other remote import path like `company.com/foo/bar` aren't supported yet.** If you want to get this implemented vote on the related issue https://github.com/OctoLinker/browser-extension/issues/264.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1299,6 +1299,7 @@ Go software and plugins.
 * [go-callvis](https://github.com/TrueFurby/go-callvis) - Visualize call graph of your Go program using dot format.
 * [go-pkg-complete](https://github.com/skelterjohn/go-pkg-complete) - Bash completion for go and wgo.
 * [go-swagger](https://github.com/go-swagger/go-swagger) - Swagger 2.0 implementation for go. Swagger is a simple yet powerful representation of your RESTful API.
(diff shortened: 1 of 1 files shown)


stefanbuck:
Thank you :smile:

kirillDanshin:
@stefanbuck waiting OctoLinker/browser-extension#264 to support my private projects.

PR #1332, opened 2017-03-21
Fix markdown (Vegeta)

reujab:
The extra space in the markdown prevented it from being rendered properly with [GitHub's new markdown engine](https://github.com/blog/2333-a-formal-spec-for-github-flavored-markdown).


kirillDanshin:
thanks

PR #1338, opened 2017-03-26
Add @cassiobotaro as new maintainer

cassiobotaro:
It's not a package contribution but it's a people aquisition.


kirillDanshin:
Welcome ;)

PR #1342, opened 2017-03-28
working on gfm compat (helps #1224)

### LINKED ISSUE #1224: Date/Time header link in TOC broken on awesome-go.com
- Date/Time header link in TOC doesn't work
- ORM list appears as a paragraph, not a list

kirillDanshin:
Signed-off-by: Kirill Danshin <k@guava.by>
helps #1224

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,8 +6,4 @@ go:
 sudo: false
 
 install:
-  - go get github.com/russross/blackfriday
-  - go get github.com/shurcooL/github_flavored_markdown
-  - go get github.com/PuerkitoBio/goquery
-  - go get github.com/gorilla/mux
-  - go get github.com/microcosm-cc/bluemonday
+  - go get -t -v ./...
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,6 +7,7 @@ sudo: false
 
 install:
   - go get github.com/russross/blackfriday
+  - go get github.com/shurcooL/github_flavored_markdown
   - go get github.com/PuerkitoBio/goquery
   - go get github.com/gorilla/mux
   - go get github.com/microcosm-cc/bluemonday
--- a/repo.go
+++ b/repo.go
@@ -8,7 +8,6 @@ import (
 	"text/template"
 
 	"github.com/gorilla/mux"
-	"github.com/microcosm-cc/bluemonday"
 	gfm "github.com/shurcooL/github_flavored_markdown"
 )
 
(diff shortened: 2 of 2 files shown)


repo.go
```diff
@@ -61,15 +42,7 @@ func generateHTML() {
 	input, _ := ioutil.ReadFile(readmePath)
 	body := string(
 		bluemonday.UGCPolicy().SanitizeBytes(
```
shurcooL: `github_flavored_markdown.Markdown` already performs sanitization, so it's not neccessary to call `bluemonday.UGCPolicy().SanitizeBytes` on its output again.

See https://github.com/shurcooL/github_flavored_markdown/blob/3c64cb3ce00a3f4feae0c53ce7c2cc2fdacba14b/main.go#L38 and https://github.com/shurcooL/github_flavored_markdown/blob/3c64cb3ce00a3f4feae0c53ce7c2cc2fdacba14b/main.go#L90-L101.

So this can be simplified to:

```Go
body := string(gfm.Markdown(input))
```

shurcooL:
I think this is a good change on a high level, because the goal is to render the Markdown on awesome-go.com in a way that is compatible with GitHub's rendering of the README.md on github.com.

That said, it doesn't look like this change would actually fix issue #1224 to me now. But that's because `github_flavored_package` also uses `sanitized_anchor_name`, the same as blackfriday's HTML renderer, so they both escape "Date & Time" to "#date-time" instead of "#date--time" as github does. But that's actually a bug in `github_flavored_package` and should be fixed there.

So, it'd make sense to rephrase this PR as "Helps #1224." or "Updates #1224." for now.

.travis.yml
```diff
@@ -7,6 +7,7 @@ sudo: false
 
 install:
   - go get github.com/russross/blackfriday
+  - go get github.com/shurcooL/github_flavored_markdown
   - go get github.com/PuerkitoBio/goquery
   - go get github.com/gorilla/mux
   - go get github.com/microcosm-cc/bluemonday
```
shurcooL: Doing `go get github.com/shurcooL/github_flavored_markdown` will implicitly get `github.com/russross/blackfriday` and `github.com/microcosm-cc/bluemonday`, because they're its dependencies. So they can be removed here.

Additionally, this entire manual list of dependencies can be simplified to just:

```
go get -t -v ./...
```

Wouldn't that be better? That's what most Go projects do, e.g., see [here](https://github.com/gorilla/mux/blob/599cba5e7b6137d46ddf58fb1765f5d928e69604/.travis.yml#L19).

kirillDanshin: @shurcooL I personally do that too, but it may be used to see project's dependencies list, so I just leave it as is.

shurcooL:
> So, it'd make sense to rephrase this PR as "Helps #1224." or "Updates #1224." for now.

I see you've done that by now. Are we ready to merge this now?

kirillDanshin:
@shurcooL yes, this patch is ready

shurcooL:
Feel free to merge then, it has my LGTM.

PR #1346, opened 2017-04-01
Include vendor to be easy to reproduce tests and examples

cassiobotaro:
Proposal to include vendor in packages to be easy to reproduce tests and examples. It's a practice large adopted by go community.


shurcooL:
With the current tooling, there isn't an easy way for _libraries_ to vendor their dependencies. It's typically done for the command that imports everything. So I don't think it's practical to ask for this from libraries.

None of my libraries vendor other libraries, they import them and I update them to work with latest version.

kirillDanshin:
Sorry, I'm :-1:. Current tools are not ready for now, so this requirement should not be merged before we have good vendoring tools that works perfectly with `go get -u` too.

Issue #1213, opened 2016-12-09
Show GitHub star count next to project link.


lishuhao:
It's great, but it will be nice display star counts in the list, sorry for my poor english

joeybloggs:
Stars are often not indicative of quality as people use them for bookmaking

kirillDanshin:
I already think about it and I think it is a bad idea.
we all know about iris a their almost-one-day 5000-star increment, and we all know that nobody will rebase their PRs every hour to solve merge issues.

shurcooL:
> and we all know that nobody will rebase their PRs every hour to solve merge issues.

Doing something like this would have to be done via an API, there's no way to manage this manually.

> Stars are often not indicative of quality

How about the number of importers (other packages that import a given package)? It'd be harder to do, but more meaningful than stars, maybe.

godoc.org has an API that exposes its knowledge of importers. E.g., the `github.com/google/go-github/github` package is currently imported by [359 packages](https://godoc.org/github.com/google/go-github/github?importers).

nouney:
With a simple python script and  [go-search](http://go-search), it's possible to do something like [that](https://github.com/nouney/awesome-go).

joeybloggs:
Unless I'm mistaken the number of importers is not guaranteed either as the numbers will not include private importers.

kirillDanshin:
@joeybloggs yep, that's right.
exemplī grātiā, I use `dlog` in about hundred packages but https://godoc.org/github.com/kirillDanshin/dlog?importers shows only one.

joeybloggs:
I don't mean to sound like a pessimist, and all of the ideas above are good, however this is a discussion has been had before not only here but in many similar sites without a concrete answer/solution

- stars are too arbitrary due to people using for bookmarking 
- importers only shows public and not private
- latest activity isn't reliable as the code may be stable
- the list goes on....

I would personally not like a unit of measurement on any package/link as it can incorrectly lead people to the conclusion that more(or less depending on metric) is better.

I would say that usually each package has something unique about it and it is up to the developer to determine which is best for his/her situation.

I would hate to see some great packages overlooked, just because it didn't have a bunch of stars.

AlekSi:
We can use @buger's expirience from building https://ossmetrics.com.

buger:
Thank you for mentioning.

[...]

PR #1368, opened 2017-04-15
Add github stars badges after github links

TrueFurby:
I have added badges after all github links in README for better overview. 

download

What do you think?

> P.S: used regular expression: https://gist.github.com/TrueFurby/444d7fc42898d98a2ad6ce67d1260083


joeybloggs:
This was/is discussed here https://github.com/avelino/awesome-go/issues/1213

TrueFurby:
Right.. sorry I didn't look for similar issue.

For anyone else who might want to see star badges use userscript like this one: https://gist.github.com/TrueFurby/46d722403b52a899ed77398688c6b0cb

shurcooL:
I'll just briefly mention that if we were to consider making this kind of change, we would need to carefully look at the performance characteristics of the implementation.

There are over 1000 repositories listed in awesome-go, which is 1000+ badges. There are thousands of people viewing the awesome-go list daily. If the implementation is inefficient (i.e., no caching or inefficient caching, etc.), we could inadvertently DDoS github's API or the badge provider.

I haven't considered the performance of the current implementation, so I can't say if this is a problem or not. I'm just saying it's something we would need to do before ever merging such a feature.

PR #1396, opened 2017-04-28
Cleanup Web Frameworks section

joeybloggs:
The Web Framework section was becoming littered with libraries, routers etc. that should not be considered "Web Frameworks", to that end some libraries have been shuffled to their appropriate sections and a new "Routers" sub-section was created to distinguish URL routers and multiplexers.

I propose that some guidelines or a checklist should be created to help determine what should be considered a "Web Framework" vs "Something Else" to avoid the section becoming cluttered again, but that can be done at a later date.

### DIFF
--- a/README.md
+++ b/README.md
@@ -60,6 +60,7 @@ Please take a quick gander at the [contribution guidelines](https://github.com/a
     - [Version Control](#version-control)
     - [Video](#video)
     - [Web Frameworks](#web-frameworks)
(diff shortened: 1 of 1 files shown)


cassiobotaro:
thanks @joeybloggs !

matrixik:
Thank you for your work.

avelino:
Thanks

PR #1400, opened 2017-05-01
recommendation new coverage runner cover.run

avelino:
Shows how many percent that covered the package


joeybloggs:
If we're going to link to `cover.run` I think it's only fair that we add links to the others suggested also.

other than that LGTM

kirillDanshin:
LGTM, bro. Feel free to merge it

PR #1440, opened 2017-05-30
Fix anchors

loopthrough:
It is a simple fix for broken anchors on first two links.


shurcooL:
Related: #1224.

This will work either on the GitHub README, or on the website, but not both, until #1224 is fully resolved.

avelino:

PR #1459, opened 2017-06-16
Add a deadline for new PR on CONTRIBUTION

PotHix:
We are trying to avoid having hundreds of open Pull Requests. By using
this rule we may close old PRs.

ref #1458


avelino:
thanks @PotHix

Issue #1446, opened 2017-06-05
Please curate this list


kweber-tcc:
This seems to me to be a list of any Go library that is submitted.  There are MANY libraries included with less than 100 stars and absolutely no documentation.  (Examples: https://github.com/siddontang/go-log, https://github.com/white-pony/go-fann, https://github.com/daviddengcn/go-pr, https://github.com/e-dard/godist)  Many libraries have no tests.  Many are no even stable yet.

What is the point of a curated list that accepts anything submitted?  How is this a list of awesome Go code?  Why are you not even adhering to the quality standards set out in your own README?

Please remove the less-than-awesome libraries, so this is actually an awesome list.

kamilsk:
good point (except `less than 100 stars`)

avelino:
I find it relevant to this question to improve the quality of the list.

Star in Repository does not demonstrate the quality of the project, today we review the proposals asking for the following items:
- test (coverage should be >= 80%, tool to help [cover.run](https://cover.run/). There's a packet we can't charge that item, example: #1594)
- lint
- documentation
read more: https://github.com/avelino/awesome-go/blob/master/CONTRIBUTING.md#quality-standards

I think it's important to ask for minimum X stars, my vote is for **42 stars** *("The answer to the ultimate question of life, the universe and everything is 42")* 

What's your opinion? @cassiobotaro @felipeweb @crgimenes @shurcooL @dukex @matrixik @joeybloggs @kirillDanshin @appleboy @PotHix @campoy @ianlancetaylor

marioidival:
I do not like this measure by stars. Tests and doc it's better to this...

caarlos0:
Maybe having a good score on goreportcard as well?

shurcooL:
> I think it's important to ask for minimum X stars, my vote is for 42 stars ("The answer to the ultimate question of life, the universe and everything is 42")
>
> What's your opinion?

Number of stars is a measure of how interesting a library is multiplied by how many people have seen it. If a good library is created and not yet shared everywhere, it won't have many stars. That doesn't mean it's not good, just that it hasn't been shared/seen by many people yet.

For example, https://github.com/golang/arch is an official Go repository that has 17 stars at this moment.

![image](https://user-images.githubusercontent.com/1924134/31511142-c93b0f2c-af54-11e7-9786-7f0eeb8b18c3.png)

I don't think there should be a minimum star requirement, because stars are not a measure of quality, they're a measure of interest and exposure.

[...]

PR #1474, opened 2017-06-28
Added v

corpix:
This is a uniform tool to manage toolchain versions. It manages only Go and Node.js versions at this time but, in theory, could be extended to manage any other language toolchain.
Goals of this tool are:
- to be simple "one file" solution
- which does not require any installation or modification of bash/zsh/*rc files
- written in pure Python 3 without any other dependencies than python and git binary

-------

- github.com repo: https://github.com/corpix/v
- godoc.org: It have no godoc, tool is implemented in pure Python without dependencies
- goreportcard.com: None

**Make sure that you've checked the boxes below before you submit PR:**
> Some items are unchecked because this is just utility, not Go package.
- [X] I have added my package in alphabetical order
- [X] I know that this package was not listed before
- [ ] I have added godoc link to the repo and to my pull request
- [ ] I have added coverage service link to the repo and to my pull request
- [ ] I have added goreportcard link to the repo and to my pull request
- [X] I have read [Contribution guidelines](https://github.com/avelino/awesome-go/blob/master/CONTRIBUTING.md#contribution-guidelines), [maintainers note](https://github.com/avelino/awesome-go/blob/master/CONTRIBUTING.md#maintainers) and [Quality standard](https://github.com/avelino/awesome-go/blob/master/CONTRIBUTING.md#quality-standard).

### DIFF
--- a/README.md
+++ b/README.md
@@ -1450,6 +1450,7 @@ See [go-hardware](https://github.com/rakyll/go-hardware) for a comprehensive lis
 * [Scaleway-cli](https://github.com/scaleway/scaleway-cli) - Manage BareMetal Servers from Command Line (as easily as with Docker).
 * [sg](https://github.com/ChristopherRabotin/sg) - Benchmarks a set of HTTP endpoints (like ab), with possibility to use the reponse code and data between each call for specific server stress based on its previous response.
 * [StatusOK](https://github.com/sanathp/statusok) - Monitor your Website and REST APIs.Get Notified through Slack, E-mail when your server is down or response time is more than expected.
+* [v](https://github.com/corpix/v) - Provides an command-line interface to manage Go versions.
 * [Vegeta](https://github.com/tsenart/vegeta) - HTTP load testing tool and library. It's over 9000!
 * [webhook](https://github.com/adnanh/webhook) - Tool which allows user to create HTTP endpoints (hooks) that execute commands on the server.
 * [Wide](https://wide.b3log.org/login) - A Web-based IDE for Teams using Golang.


cassiobotaro:
@avelino the tool is written in python, but can be useful in go, what you thing about it?

corpix:
@avelino maybe you could show me what's wrong?

kirillDanshin:
LBTM. we even didn't add travis/gimme. I think we don't need to add it too.

corpix:
@kirillDanshin Could you please explain me why? «we even didn't add travis/gimme» sounds as unconstructive criticism to me.
I think this tool could be helpful for developers who use go to test the code with different versions of the toolchain. You have `gvm` in the list, so why not add a tool which has an alternative point of view on problems which `gvm` solves?

geovanisouza92:
@corpix Why it's written in Python instead of Go?

corpix:
@geovanisouza92 because:

1. it exists in almost every linux distribution
2. this tool is intended to be a solution which does not require a system with pre-installed go toolchain
3. I don't like an idea of "binary distribution" for the software because I don't trust other people and don't want others to trust me(python code is transparent and could be easily reviewed by community in contrast to binary release)

crgimenes:
It does not look good to me either.
Consider rewriting in Go.

corpix:
You don't like a tool just because it is not written in Go. Okay, I got the point.

crgimenes:
Do not get me wrong, it seems like a great tool. But the list is about Go, not about Python. We could not revise the code, and apply the same pattern we applied to the other PRs, we would have to open many exceptions, it would not be fair to others.

corpix:
Yes, this will be unfair. Well, rules are rules, this is ok.
But following this rule some tools which are already in list should be rejected because their repositories does not contain a Go code:

> `gobrew`, `gvm` and `rodent` solves a similar set of problems `v` is trying to solve

- https://github.com/c9s/c6
- https://github.com/Mainflux/mainflux
- https://github.com/VividCortex/johnny-deps
- https://github.com/rjohnsondev/vim-compiler-go
- https://github.com/Microsoft/vscode-go
- https://github.com/skelterjohn/go-pkg-complete
- https://github.com/cryptojuice/gobrew
- https://github.com/moovweb/gvm
- https://github.com/alouche/rodent

PotHix:
@corpix You're right on this one. There is an issue for almost the same purpose (#1446), please comment there about this. 

Please provide separate PR's with your arguments to remove each of the libs mentioned. :)

corpix:
@PotHix I don't support an idea to get rid of the things written in programming languages other than Go(because they could be useful for other people), so I won't, sorry :)

PotHix:
No problem at all, I was just following your own argument: `But following this rule some tools which are already in list should be rejected because their repositories does not contain a Go code`. :)

PR #1505, opened 2017-07-27
Broken anchor links

joelterry:
A few anchor links were broken, likely since https://github.com/shurcooL/sanitized_anchor_name/commit/f105e6d6adf3fd73b316934472e2a3c0146fb3a9


joelterry:
Should we use Github's markdown API (https://developer.github.com/v3/markdown) instead of github.com/shurcooL/github_flavored_markdown, since there's a discrepancy in how they generate links?

avelino:
Today generate HTML in https://github.com/avelino/awesome-go/blob/master/repo.go#L36
You send PR using github api?

joelterry:
Here's the Github code that I believe is causing the weird anchor links: 

https://github.com/jch/html-pipeline/blob/ca2c47846af141f6b217a65420b8adefda81217c/lib/html/pipeline/toc_filter.rb

I might submit a pull request or raise an issue with them, but in the short-term, using an "and" instead of an "&" would work (as mentioned in https://github.com/avelino/awesome-go/issues/1224). https://github.com/vinta/awesome-python does this, btw.

cassiobotaro:
I like this "and" solution, can you take a look @avelino ?

kirillDanshin:
+1. LGTM. I think we should merge this fix and then work on making `&` available in titles.

PR #1562, opened 2017-08-29
Remove deprecated thing

hvnsweeting:


avelino:
Where do you describe it was depreciated?
I didn't see it https://github.com/go-lang-plugin-org/go-lang-idea-plugin

hvnsweeting:
From what was written in the line 

```
 -* [go-lang-idea-plugin](https://github.com/go-lang-plugin-org/go-lang-idea-plugin) (deprecated) - The previous Go plugin for IntelliJ (JetBrains) IDEA, now replaced by the official plugin (above). 
```

cassiobotaro:
"Please note, that the following features are available in Gogland, JetBrains's official IDE, or JetBrains maintained plugin, but not in this plugin". I think we can remove deprecated plugin.

PR #1583, opened 2017-09-13
Add go-on-rails as a Web Framework

gingerhot:
Add [go-on-rails](https://github.com/goonr/go-on-rails) to the list under the `Web Frameworks` node.


gingerhot:
In fact go-on-rails is a Rails generator helping building Go APIs, so there's no godoc and goreportcard.com links. But I think it's relative to Go programming especially to those Rails developers who also do Go programming.

If it really not belongs to this category, so let it be will be ok.

Thanks~

cassiobotaro:
@gingerhot Thanks for submit this tool, but unfortunately is not a "awesome go" tool, but a ruby tool. Can be useful, but still not made by go.

PR #1638, opened 2017-10-20
Remove go.auth

flowonyx:
https://github.com/bradrydzewski/go.auth says in the description that it is deprecated and it has not been updated since 2013.


cassiobotaro:
thanks @flowonyx

PR #1683, opened 2017-11-15
Add 1backend, remove borg

crufter:
I have removed Borg, which I have submitted a year ago but I don't maintain anymore.
I have added 1Backend, a PaaS service which has a go backend.

- github.com repo: https://github.com/1backend/1backend
- godoc: https://godoc.org/github.com/1backend/1backend


crufter:
@avelino I havent added those links because for a standalone app like this I'm not sure if it makes sense:
- gocover: We do testing with end-to-end tests, so no or very few tests
- godoc: there is no Go API to document just HTTP endpoints

That might sound bad but there is a lot of love and effort going into the project :D

PR #1702, opened 2017-11-28
Add distillog to 'Logging' section

amoghe:
Adds [distillog](https://github.com/amoghe/distillog) to the `Logging` section.


amoghe:
This seems to be missing the standard PR template. Will recreate the PR with the template.

PR #1707, opened 2017-11-29
Better slack badge

cassiobotaro:
Better slack badge instead of raw text.


peergynt:
It looks like the slack badge is now broken

PR #1774, opened 2018-01-26
Reference forked migrate package.

dhui:
Old repo is no longer maintained: https://github.com/mattes/migrate/issues/311
Addresses: https://github.com/golang-migrate/migrate/issues/5


jtemporal:
Great to see that you are continuing `migrate` but I would only merge this once the organization is "more populated" as mentioned [here](https://github.com/mattes/migrate/issues/311#issuecomment-360889078) **and** the coverage increases. Today's coverage doesn't meet the quality standards on our contributing guide:
>Tests, where practical. If the library/program is testable, then coverage should be >= 80% for non-data-related packages and >=90% for data related packages.

avelino:
It is necessary to await the official decision of the @mattes

PR #1803, opened 2018-02-12
Fix go-finance bad markdown syntax

Southclaws:
(just a tiny fix so I figured the PR template didn't really apply to this situation!)


kirillDanshin:
Thanks! :)
The part that you're awesome _is_ really apply ;)

PR #1910, opened 2018-05-01
add cmp - Provides comparison for different Go types and a Comparer i…

mickep76:
add cmp - Provides comparison for different Go types and a Comparer interface.

[GoDoc](https://godoc.org/github.com/mickep76/cmp)
[codecov](https://codecov.io/gh/mickep76/cmp)
[Build Status](https://travis-ci.org/mickep76/cmp)
[Go Report Card](https://goreportcard.com/report/github.com/mickep76/cmp)
[License](https://github.com/mickep76/mlfmt/blob/master/LICENSE)

### DIFF
--- a/README.md
+++ b/README.md
@@ -257,6 +257,7 @@ Please take a quick gander at the [contribution guidelines](https://github.com/a
 * [bloom](https://github.com/zhenjl/bloom) - Bloom filters implemented in Go.
 * [bloom](https://github.com/yourbasic/bloom) - Golang Bloom filter implementation.
 * [boomfilters](https://github.com/tylertreat/BoomFilters) - Probabilistic data structures for processing continuous, unbounded streams.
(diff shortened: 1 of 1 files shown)


mickep76:
Go lint is fixable but they are from the example not the actual code, go cyclo not fixable since the function needs to be somewhat larger then what is normal.

mickep76:
Fixed the lint errors in the example.

mickep76:
Thought a little about, I could reduce the switch statement to using a Map with individual funcs for each comparison like:

```
func EqInt8(a, b interface{}) bool {
        return a.(int8) == b.(int8)
}

var EqMap map[reflect.Type]func(a, b interface{}) bool {
        reflect.Int8: EqInt8,
}

func Eq(a, b interface{}) (bool, error) {
        f, ok := EqMap[t] ...
}
```

Not sure that's an improvement but it would be one way of doing it.

avelino:
`gocyclo 50%`

PR #2073, opened 2018-08-19
Add Lazygit

sherry-ummen:
Its a nice little tool. Looks promising, I think it should be in this list to give more recognition


sherry-ummen:
I will make the pull request again

PR #2126, opened 2018-10-03
Fix goreportcard URL

BattleBas:
Fixed the goreportcard URL by removing the "report" part which was leading to a 404 page. Now the URL points to the homepage, I assume this is what was intended.


BattleBas:
Ah ok, so the documentation should then be updated to "https://goreportcard.com/report/github.com/ github_user / github_repo" then? I was confused because the link was clickable I thought you intended it to be followed. Thanks!

PR #2146, opened 2018-10-11
Change tsdb link of Prometheus

codesome:
Currently under `Database` section, `prometheus` is not actually pointing to the time series database, but to the monitoring system.

This PR replaces it with the actual Time Series Database repo of Prometheus.


codesome:
After discussing with Prometheus maintainers, apparently the current link should be fine.

PR #2273, opened 2018-12-06
Add gonum

ceriath:
Follow-up for #2258 

I don't really like https://goreportcard.com/report/github.com/gonum/gonum, but seems like mostly generated code and the usefulness of the package might justify adding it


cassiobotaro:
@ceriath the package description should finish with a period.

PR #2294, opened 2018-12-21
Google Drive client is not a cli parser library

artburkart:
I think the Google Drive client was added to the wrong place in the README. This PR fixes that.


ceriath:
@artburkart thanks! regarding uptimerobot, it is technically a library, but they also provice a CLI, so i think it is fine to list it as library

artburkart:
:shipit: 👍

Issue #1749, opened 2018-01-03
awesome-go badge


ldemailly:
related to #1309 but slightly different:

Have a specific documented markdown would be great,
for now
```
[![Awesome-Go](https://cdn.rawgit.com/sindresorhus/awesome/d7305f38d29fed78fa85652e3a63e154dd8e8829/media/badge.svg)](https://github.com/avelino/awesome-go)
# or https://awesome-go.com/ ?
```
maybe ?

ldemailly:
or a variant of 
https://raw.githubusercontent.com/sindresorhus/awesome/master/media/mentioned-badge.svg

ldemailly:
I used
![](https://raw.githubusercontent.com/istio/fortio/master/docs/mentioned-badge.svg?sanitize=true)

avelino:
Sorry, I did not understand this issue, what would be your proposal?

ldemailly:
Publish a badge and a snippet of GitHub markdown that projects can add to their Readme once listed

avelino:
Does everyone like it?
My proporsal: ![awesome-go](https://user-images.githubusercontent.com/31996/36610603-957fabd8-18af-11e8-89be-41b1873c8de1.png)

avelino:
We're going with this one?
[![Awesome-Go](https://cdn.rawgit.com/sindresorhus/awesome/d7305f38d29fed78fa85652e3a63e154dd8e8829/media/badge.svg)](https://github.com/avelino/awesome-go)

We need an automated test to check the README to see if it contains the badge. Who's proposing?
@ldemailly

ldemailly:
I don't think automated test is needed, just have people have the option to use badge is probably fine. ie point to it in https://github.com/avelino/awesome-go/blob/master/README.md
with the markdown to use, something like
```
[![Awesome Go](https://raw.githubusercontent.com/avelino/awesome-go/master/docs/mentioned-badge.svg?sanitize=true)](https://github.com/avelino/awesome-go#YOURSECTION)
```

assuming you check in the badge as svg in docs/mentioned-badge.svg

ldemailly:
can you please un-assign me from this issue, I'm cleaning up my assigned issues

ceriath:
closing this since #2295 got merged

PR #2295, opened 2018-12-21
 Increase response time and add mentioned badge

ceriath:
Regarding #1749:
https://github.com/sindresorhus/awesome/blob/master/awesome.md#awesome-mentioned-badge is pretty clear on how it works. This encourages users to add the badge.

I also increased the response time to 30 days, since i think thats what we are doing right now anyways


avelino:
very very 30 recommendation 😄

PR #2296, opened 2018-12-21
Revert "Added gqlc "GraphQL Compiler""

ceriath:
See #2288


ceriath:
woops, this would revert #2285 too

PR #2297, opened 2018-12-21
Remove gqlc

ceriath:
See #2288


kirillDanshin:
I think @Zaba505 should send us a PR to add gqlc as soon as it will do everything stated in the docs.

PR #2313, opened 2019-01-07
Disambiguate listed flac libraries

hunterwerlla:
- Issue: the two listed flac libraries are named the same and have the same description
- Added a more detailed description that makes it clear the two libs do not have the same capabilities


hunterwerlla:
Oops, fixed the spelling mistakes

PR #2429, opened 2019-03-16
Add gophercises as a resource

shantanu404:
Gophercises is a great platform for new programmers in the parade who are already familiar with another language but new to Go. It takes a _DYI_ approach which I think is absolutely fantastic! I think it will help others as it helped me :heart: 

Note that I am adding this as a resource under _websites_ so I didn't use the normal PR template which was more geared to new projects.


shantanu404:
:+1:

Issue #2440, opened 2019-03-22
https://awesome-go.com/ shows a Watch Movies website


complex64:
https://awesome-go.com/ shows:

![2019-03-22-224920_1047x598_scrot](https://user-images.githubusercontent.com/63771/54854928-cc032480-4cf4-11e9-8477-5480fdbd71dd.png)

kirillDanshin:
@avelino? O_o

avelino:
sorry

complex64:
No worries, thanks for the fix @avelino

PR #2441, opened 2019-03-23
Netlify hosting

avelino:
Digitalocean server has been shut down by sponsor, migrating to Netlify (free plan)
fixed: #2440


complex64:
In case this becomes an issue again, we'd be happy to sponsor something in the future

kirillDanshin:
I’m also can be a sponsor, @avelino can ping me too when awesome-go will need a sponsor. Or, we can also setup a patreon campaign. Choice is up to @avelino

PR #2462, opened 2019-04-08
Update torrent entry

anacrolix:


ceriath:
Please provide links and fill in the PR form.

anacrolix:
That was intentional. It's a related project, but not a library. I think it was an early adopter when the torrent repo was first added to this list.

PR #2631, opened 2019-07-20
add performance section

kamilsk:
changes:
- moved jaeger, profile, and tracer to the new section Performance
- removed go-torch because it's archived
- removed netbug because it doesn't have any tests


kamilsk:
friendly reminder

PR #2638, opened 2019-07-27
Add GopherGameServer

hewiefreeman:
I'm not quite sure how to cover most of this project with useful tests, as almost all of the functionality requires the server to be running (with a client or two connected), or to be connected to a MySQL database. Though, all features have been extensively tested, and the project is very stable and well documented.

- github.com repo: https://github.com/hewiefreeman/GopherGameServer
- godoc.org: https://godoc.org/github.com/hewiefreeman/GopherGameServer
- goreportcard.com: https://goreportcard.com/report/github.com/hewiefreeman/GopherGameServer
- coverage: N/A (?)

PS: If test coverage is absolutely needed, I'll try my best to cover everything that is reproducible without developers (or you) having to modify the test files (i.e. database IP/ports, localhost port, etc) or do extra non-Go setup for it to pass.

### DIFF
--- a/README.md
+++ b/README.md
@@ -698,8 +698,8 @@ Please take a quick gander at the [contribution guidelines](https://github.com/a
 * [go-collada](https://github.com/GlenKelley/go-collada) - Go package for working with the Collada file format.
 * [go-sdl2](https://github.com/veandco/go-sdl2) - Go bindings for the [Simple DirectMedia Layer](https://www.libsdl.org/).
 * [go3d](https://github.com/ungerik/go3d) - Performance oriented 2D/3D math package for Go.
-* [GopherGameServer](https://github.com/hewiefreeman/GopherGameServer) - Full-featured, and easy to use Go game server with a JavaScript client API.
 * [gonet](https://github.com/xtaci/gonet) - Game server skeleton implemented with golang.
+* [GopherGameServer](https://github.com/hewiefreeman/GopherGameServer) - Full-featured, and easy to use Go game server with a JavaScript client API.
 * [goworld](https://github.com/xiaonanln/goworld) - Scalable game server engine, featuring space-entity framework and hot-swapping.
 * [Leaf](https://github.com/name5566/leaf) - Lightweight game server framework.
 * [nano](https://github.com/lonng/nano) - Lightweight, facility, high performance golang based game server framework.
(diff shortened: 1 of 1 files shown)


hewiefreeman:
@ceriath thank you for the warm review and pointing that out!

I actually thought that ineffassign was a false-positive for a very long time after I had revised and improved that custom hand-shake it's in. But with further investigation today, it in fact is an ineffectual assignment. Fixed, and am committing the changes for review.

kirillDanshin:
@hewiefreeman Sorry for delay. I really like the project, but I see that you really need to work on gocyclo warnings. For example, see https://github.com/hewiefreeman/GopherGameServer/blob/master/database/setup.go#L9

Also, please use conventional Go comments format: here ( https://github.com/hewiefreeman/GopherGameServer/blob/master/database/setup.go#L8) you can see all-caps, which is pretty hard to read at scale

Closing this one as I'm cleaning up our PRs. Feel free to ping me in the new PR.

Please, open a new PR as soon as you done.

hewiefreeman:
@kirillDanshin Thanks for taking a look! I'll go through the code and replace any all-caps comments. 

Though, I'm not quite sure what to do about the gocyclo warnings, since I thought a function call is a lot more expensive than in-lining instructions? Would the go compiler inline everything for me if I were to break those functions down into smaller ones? I do like the idea of making a function as small and readable as possible, but this is a server, so why should I compensate readability over execution/compile speed?

kirillDanshin:
@hewiefreeman you can take a look in the fasthttp sources for examples. we're trying to keep things small, and we still keep the performance.

If we're talking about performance, normally if you compare a huge chunk of code in some `if` statement with a function call, you can notice that they're usually about the same thing from a performance perspective.

you can read further info here to know more about how the `if` statement works: https://www.intel.com/content/dam/www/public/us/en/documents/manuals/64-ia-32-architectures-software-developer-instruction-set-reference-manual-325383.pdf on page 486

Speaking of main() or database setup, those functions doesn't need to be so fast, so you think about manually inlining your functions. In this way you'll just get a program that is hard to maintain.

hewiefreeman:
What would you suggest for complicated checks such as https://github.com/hewiefreeman/GopherGameServer/blob/master/server.go#L231 ? I broke that off from the Start() function, but I still don't understand what I should be using instead of an if statement when I need a check, or when I want to group relevant checks such as that example.

kirillDanshin:
I'd suggest to split this checks into different functions like `verifyDBSettings` etc.

also, you can omit `== true` in conditions to make your code easier to read

PR #2657, opened 2019-08-03
Remove cosiner/zerver

ss7loginov:
404 Not Found


kirillDanshin:
I found zerver clone here https://github.com/leeonsoft/zerver, but anyway it was not updated for 4 years

PR #2671, opened 2019-08-11
Add m-zajac/json2go

m-zajac:
Added https://m-zajac.github.io/json2go

I'd like to propose adding this little utility that I think is pretty convenient for go community :) 
It's a bit different than existing tools, especially works well when provided an array of docs that are not exactly the same.

Code is written purely in go, and web page uses webassembly created by go compiler. 
I've tried my best to ensure good code quality and test coverage.

Few examples to show what this tool can do can be found here: https://github.com/m-zajac/json2go#example-outputs

Thanks for maintaining this awesome repository :)


kirillDanshin:
github pages link should be placed under websites category, while the link to the library itself can be placed in this category. Please, move the link and open a new PR then.
Thanks.

P.S. nice project tho

PR #2683, opened 2019-08-18
Removing rkt: it has been archived by the CNCF :(

clockworksoul:
This is a sad day: `rkt` has been [officially archived by the CNCF](https://www.cncf.io/blog/2019/08/16/cncf-archives-the-rkt-project/). 

It probably makes sense to remove it from this list. 😢 This diff does exactly that.

### DIFF
--- a/README.md
+++ b/README.md
@@ -2122,7 +2122,6 @@ See [go-hardware](https://github.com/rakyll/go-hardware) for a comprehensive lis
 * [peg](https://github.com/pointlander/peg) - Peg, Parsing Expression Grammar, is an implementation of a Packrat parser generator.
 * [Pipe](https://github.com/b3log/pipe) - A small and beautiful blogging platform.
 * [restic](https://github.com/restic/restic) - De-duplicating backup program.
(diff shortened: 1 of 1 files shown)


kirillDanshin:
`rkt` will keep existing as a standalone free-software project, but it won't be anymore under CNCF umbrella.

ref: https://github.com/rkt/rkt/issues/4004#issuecomment-507358362

I don't think that losing CNCF umbrella can suddenly make an awesome project be a bad project.

clockworksoul:
Perhaps, but it's also effectively dead, which is why it was archived. 

Genuine question: does this list have a stance on listing inactive projects?

kirillDanshin:
@clockworksoul

actually, I think we should remove `rkt`, but clearly state that we're not removing it due to losing a CNCF umbrella, but due to the fact it wasn't maintained for a long time.

I would be glad to see the project revived at some point later.

> Genuine question: does this list have a stance on listing inactive projects?

no, I just think it's fair to notice, that `rkt` still have a chance for active development, and I hope it will be again in our list 🙂

clockworksoul:
I don't have the access to merge the approved change. Who takes care of that?

PR #2749, opened 2019-10-01
Create .dockerignore to speedup build.

hloeffler:
Create .dockerignore to speedup build.

I just want to stop docker from reading ".git".


avelino:

PR #2787, opened 2019-10-19
Removed unmaintained abutil

magmast:
abutil package is no longer maintained, so it should be deleted from the list.


avelino:

- godoc.org: False
- goreportcard.com: False
- coverage: False

PR #2833, opened 2019-11-24
update description

pjebs:
These are my repos. The descriptions need to be updated.


avelino:

- godoc.org: False
- goreportcard.com: False
- coverage: False

PR #2902, opened 2020-01-11
Update Fyne description to match new release info

andydotxyz:
Just a quick change to an existing project entry.


avelino:

- godoc.org: False
- goreportcard.com: False
- coverage: False

PR #3036, opened 2020-04-06
- add igo (github.com/rocketlaunchr/igo)

pjebs:
- github.com repo: https://github.com/rocketlaunchr/igo
- pkg.go.dev: N/A
- goreportcard.com: N/A
- coverage service link N/A

`igo` is a transpiler that converts igo files (go files with some new syntax and other language features) into standard go files which you can then compile.

**It is not a package for others to use.**

There is a unit test provided for converting a "sample" igo file to a standard go file which you can then then run. Peruse/run the 1 unit test in the parent directory.


avelino:

- goreportcard.com: False
- coverage: False

PR #3132, opened 2020-06-18
Quick typo fix

RobusGauli:
Just a quick typo fix


avelino:

- goreportcard.com: False
- coverage: False

PR #3143, opened 2020-06-28
Update the link address and description of rod

ysmood:
We migrated the project from personal namespace to the organization namespace.


avelino:

- goreportcard.com: False
- coverage: False

PR #3154, opened 2020-07-04
Add last-commit badges to all github repos

kpym:
Some of the projects are no longer being maintained, others are very active. The `last-commit` badge makes it easy to distinguish them.

### DIFF
--- a/README.md
+++ b/README.md
@@ -115,84 +115,84 @@ Please take a quick gander at the [contribution guidelines](https://github.com/a
 
 *Libraries for manipulating audio.*
 
-* [EasyMIDI](https://github.com/algoGuy/EasyMIDI) - EasyMidi is a simple and reliable library for working with standard midi file (SMF).
-* [flac](https://github.com/mewkiz/flac) - Native Go FLAC encoder/decoder with support for FLAC streams.
-* [gaad](https://github.com/Comcast/gaad) - Native Go AAC bitstream parser.
-* [go-sox](https://github.com/krig/go-sox) - libsox bindings for go.
-* [go_mediainfo](https://github.com/zhulik/go_mediainfo) - libmediainfo bindings for go.
-* [gosamplerate](https://github.com/dh1tw/gosamplerate) - libsamplerate bindings for go.
-* [id3v2](https://github.com/bogem/id3v2) - Fast and powerful ID3 decoding and encodung library written completely in Go.
-* [malgo](https://github.com/gen2brain/malgo) - Mini audio library.
-* [minimp3](https://github.com/tosone/minimp3) - Lightweight MP3 decoder library.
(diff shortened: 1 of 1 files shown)


avelino:

- goreportcard.com: False
- coverage: False

avelino:
This will generate a lot of demand in shields.io at each access, it's not healthy for the project.

Did you make this change via script?
This information may be interesting on awesome-go.com, but not on readme.md

panjf2000:
Indeed, adding this kind of badge will result in thousands of additional HTTP requests when visiting awesome-go page, which badly slows down the access speed.

kpym:
I've made this by simple search/replace with regular expressions. And I agree that this is not optimal. 

May be fetching all badges once a day is enough. And store them in a unique sprite image will limit the http access.

panjf2000:
The idea of checking out the activities of repositories in awesome-go list makes great sense, maybe we set up a daily script to scan all repositories in README and get the last commit, then write these last-commit dates into README back, as text format, it doesn't have to be as image format.

kpym:
@panjf2000 Yes it could be text format, but the color is very important IMO because it helps to quickly check how active is the project.

panjf2000:
> @panjf2000 Yes it could be text format, but the color is very important IMO because it helps to quickly check how active is the project.

Make sense, but we have a long way to go if the final decision is image format, then where should we store those cached images?

kpym:
The `last-commit` is only one of the useful informations that can be fetched from the github API, you can at the same time fetch the number of stars, the number of downloads, ... directly from GitHub (without shield.io). All this once a day and make it available in https://awesome-go.com.

avelino:
My proposal would be to start with text and then evolve to image (since we need to evolve the idea of where to host the images), baby steps.

@kpym can you work on this implementation? I imagine this script (in **Go**) stay inside `scripts` _folder_ in the awesome-go repository and run automatically via Github Actions

## about awesome-go.com
is hosted on netlify, that is, after PR is accepted travis-ci generates the site (html) and sends to netlify, today everything should be contained in the repository awesome-go

thinking about using images we need to improve this architecture, I believe that object store (aka s3) can help us to host the images generated

kpym:
@avelino I'm sorry, but I won't have time to deal with this right now.

Some remarks: 

- I don't think images are necessary for awesome-go.com where you can use arbitrary css to style text elements. 
- Images are only useful for rendering the README on github, but as already mentioned this is far from optimal.
- On second thought I think the README should be kept as simple as possible and that awesome-go.com should be enriched with informations about git repositories (styled via css), as well as search filters (e.g. to show projects that have more than 5 stars and that were updated less than 3 months ago).

This way awesome-go.com will have a real added value on the README, because currently this is not the case, IMO.

avelino:
I agree, we need to improve awesome-go.com

_Too bad you can't contribute awsome-go, we'll always be open to new ideas, when you have time to contribute be very welcome_

PR #3178, opened 2020-07-22
Style Guides: add section from dgryski/awesome-go-style

dgryski:
As discussed, the style guides from dgryski/awesome-go-style.


avelino:

- goreportcard.com: False
- coverage: False

dgryski:
Repushed after fixing sorting of entries.

avelino:
thanks @dgryski

PR #3210, opened 2020-08-12
WIP: test stale repositories

mrKappen:
ref: #3211


avelino:

- goreportcard.com: False
- coverage: True

test_stale_repositories.go
```diff
@@ -0,0 +1,108 @@
...
+		defer resp.Body.Close()
+		json.NewDecoder(resp.Body).Decode(&respObj)
+		isAged = len(respObj) == 0
+		if isAged {
+			body := &issue{
+				Title: fmt.Sprintf(issueFmt, ownerRepo),
+			}
+			buf := new(bytes.Buffer)
+			json.NewEncoder(buf).Encode(body)
+			req, err := http.NewRequest("POST", githubPOSTISSUES, buf)
```
avelino, line 79: if the issue already exists we cannot reopen, this can generate many repeated issues (flood)
we should see if the issue already exists and is open (looking for the title pattern)

test_stale_repositories.go
```diff
@@ -0,0 +1,108 @@
...
+	"os"
+	"regexp"
+	"strings"
+	"time"
+
+	"github.com/PuerkitoBio/goquery"
+	"golang.org/x/oauth2"
+)
+
+var reGithubRepo = regexp.MustCompile("https://github.com/[a-zA-Z0-9-.]+/[a-zA-Z0-9-.]+$")
```
avelino, line 18: has a repository that can be outside the github, example:
- gitlab
- bitbucket
- ...

scripts.go
```diff
@@ -0,0 +1,29 @@
...
+	"github.com/PuerkitoBio/goquery"
+	"github.com/russross/blackfriday"
+)
+
+func readme() []byte {
+	input, err := ioutil.ReadFile("./README.md")
+	if err != nil {
+		panic(err)
+	}
+	html := append([]byte("<body>"), blackfriday.MarkdownCommon(input)...)
```
avelino, line 16:
```go
html := fmt.Sprintf("<body>%s</body>", blackfriday.MarkdownCommon(input))
```

test_stale_repositories.go
mrKappen: hey @avelino. That is a good point. Should we restrict this to a set of possible sites (i.e. github, gitlab and bitbucket). I made use of the github api to check the commit age of a repository and there is an api. If we are handling all possible repository hosting sites how would we handle each specific one?

test_stale_repositories.go
mrKappen: @avelino I added a check in the latest commit

mrKappen:
Hey @avelino, currently there are 655 repositories that could potentially be stale according to the current definition (> 1 year since last commit). I don't want to create 655 issues in one go, so I have added a limit to the number of issues that are made each time the script is run (currently set at 10)

avelino:
can we create one issue per week containing all the repositories in the issue body with checklist example:

## repository with more than 1 year without update
- [ ] https://github.com/...
- [ ] https://github.com/...
- [ ] ...


but it has a repository with more than a year without updating that must be maintained (quality packages), for this it is ideal to have a white list to not appear in all issues

mrKappen:
> can we create one issue per week containing all the repositories in the issue body with checklist example:

hmm. When you say create an issue once a week, will the script be constantly running (never exiting) and simply send a request to create an issue every week? What I had in mind was that there would be a job that runs every week or so that executes that whole script.

avelino:
`cronjob` (like github actions) executed 1 times a week so as not to generate too many issues and end up being unfeasible to review

mrKappen:
@avelino I updated the script to create a single issue each time the script is run with a checklist of all flagged repo links. Only new repositories (defined as not being referenced in a currently open issue is added)

mrKappen:
@avelino check for dead links added as well

mrKappen:
Hi @avelino I added a workflow, but I seem unable to test it on my fork. The action does not appear on github's actions tab. Is there a requirement that the yaml file has to be on the master branch?

.github/workflows/run-check.yaml
```diff
@@ -0,0 +1,13 @@
...
+on:
+  schedule:
+    - cron: '0 * * * *'
+jobs:
+  build:
+    name: Running test
+    runs-on: windows-latest
+    strategy:
+      matrix:
+       go: ['1.12']
```
avelino, line 11: used go version `1.15.x`

avelino (changes requested):

.github/workflows/run-check.yaml
```diff
@@ -0,0 +1,13 @@
+name: Check For Stale Repositories
+on:
+  schedule:
+    - cron: '0 * * * *'
+jobs:
+  build:
+    name: Running test
+    runs-on: windows-latest
```
avelino, line 8: why run on windows?

test_stale_repositories.go
```diff
@@ -0,0 +1,233 @@
...
+			*staleRepos = append(*staleRepos, href)
+			ctr++
+		}
+	}
+}
+func testStaleRepository() {
+	query := startQuery()
+	var staleRepos []string
+	addressedRepositories := make(map[string]bool)
+	oauth := os.Getenv("GITHUB_OAUTH_TOKEN")
```
avelino, line 198: as variable environment setting in github workflow `GITHUB_OAUTH_TOKEN`

.github/workflows/run-check.yaml
mrKappen: changed to ubuntu-latest

test_stale_repositories.go
mrKappen: added environment variable OAUTH_TOKEN. @avelino, you need to create a new OAUTH token and create a secret for this repository with the same name

mrKappen:
Hey @avelino, sorry for the delay. I have added a check for archived repositories.

Jason-123-cyber (commented):
K

PR #3428, opened 2021-01-03
Update mangos

fgm:
Not using the template, as this is an update PR, not an add.

The current link redirects to the legacy archived Mangos v1: https://github.com/nanomsg/mangos-v1

It should be updated to the current v2 repo https://github.com/nanomsg/mangos


avelino:

- goreportcard.com: False
- coverage: False

PR #3562, opened 2021-04-09
fixed warnings on travisci config

avelino:
- root: deprecated key sudo (The key `sudo` has no effect anymore.)
- deploy: deprecated key skip_cleanup (not supported in dpl v2, use cleanup)

Screen Shot 2021-04-09 at 09 58 22


avelino:

- goreportcard.com: False
- coverage: False

PR #3682, opened 2021-07-28
Remove whiteshtef/clockwork

whiteShtef:
Hi!

I'd like to remove a library that I added a couple of years ago because it has gone stale and I don't actively maintain it anymore.

By the way, thanks a lot for awesome-go!


avelino:

- goreportcard.com: False
- coverage: False

PR #3908, opened 2021-11-20
Add ebook 'The Power of Go: Tools'

bitfield:
Thanks for `awesome-go`; may this PR increase its awesomeness, by just a little.


avelino:

- goreportcard.com: False
- coverage: False

bitfield:
Here is some recent activity to stop the bot automatically marking the PR as closed!

bitfield:
Still not stale.

avelino (changes requested):
There is no point in this book getting on the list, it is a paid book
If we accept this contribution we would have to list all the books of go

ref: https://github.com/avelino/awesome-go/pull/3701#pullrequestreview-745617902

bitfield:
> we would have to list all the books of go

Sorry, I thought that was the point! I didn't see anywhere that your list only covers free books.

If that's the case, then I've opened #3952 to remove the existing non-free books from the list.

PR #3952, opened 2021-12-21
Remove non-free books per policy

bitfield:
See #3701 and #3908: only free books should be listed here.

### DIFF
--- a/README.md
+++ b/README.md
@@ -2902,14 +2902,10 @@ _**Unofficial** set of patterns for structuring projects._
 
 ## E-Books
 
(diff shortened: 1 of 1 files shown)


avelino:

- goreportcard.com: False
- coverage: False

phanirithvij:
@avelino I think a paid books section makes sense but personally I'm of the same opinion as you.

Or maybe we could put up issue/discussion for a vote to see if the community wants it?

avelino:
> @avelino I think a paid books section makes sense but personally I'm of the same opinion as you.
> 
> Or maybe we could put up issue/discussion for a vote to see if the community (including the maintainers) are interested in it?

I created an issue to discuss this subject, I will post my opinion there

adebobo19:
True

PR #4080, opened 2022-03-08
Removing Damsel.

spekary:
Damsel appears to be abandoned. The most recent commit was 6 years ago.

While it appears stable, it does not correspond to current awesome-go standards. In particular, it does not have a go.mod file nor any published releases, no report card and no code coverage report.

@dskinner, please reply if you would like to keep Damsel on awesome-go.


avelino:

- goreportcard.com: False
- coverage: False

spekary:
I think we are supposed to wait a week before committing to see if there is a reply. I should probably have put the date on there.

avelino:
> I think we are supposed to wait a week before committing to see if there is a reply. I should probably have put the date on there.

I will wait for the other PRs

dskinner:
I am ok with damsel not being on awesome go. I don't think projects like damsel are practical in production so in that sense it is abandoned.

This was the first thing I wrote in Go porting it from my original write in python and it only has historical significance to me. I would actively discourage use beyond reading the code.

PR #4083, opened 2022-03-08
Removing hero

spekary:
Hero appears to be adandoned:
- The most recent commit was 2 years ago.
- It is not passing its continuous integration test.
- It is still at v0.0.2 and so is not progressing to release.
- There are many open issues.
- It does not conform to current awesome-go quality standards.

@shiyanhui, please respond if you would like to maintain hero on awesome-go.


avelino:

- goreportcard.com: False
- coverage: False

PR #4085, opened 2022-03-08
Removing Mustache

spekary:
Mustache appears to be abandoned.
- Most recent commit was 6 years ago.
- It does not have a go.mod file
- It does not conform to current awesome-go quality standards

@hoisie, please reply if you would like to keep mustache on awesome-go


avelino:

- goreportcard.com: False
- coverage: False

PR #4090, opened 2022-03-09
Reorganizing text processors

spekary:
This is a reorganization of the Text Processing category into sub-categories to make it easier to peruse.

Also removed the "inject" package since the author officially archived the project and is no longer supporting it. All other packages remain.

I reviewed all of the packages in this category, and while most don't fit current quality standards, most of them are simply mature projects with 1 or 2 open issues, and are still quite useful.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1594,8 +1594,7 @@ _Libraries for working with human languages._
 
 See also [Text Processing](#text-processing).
 
-### Case Converters
-- [kace](https://github.com/codemodus/kace) - Common case conversions covering common initialisms.
+### Slugifiers
 - [go-slugify](https://github.com/mozillazg/go-slugify) - Make pretty slug with multiple languages support.
 - [slug](https://github.com/gosimple/slug) - URL-friendly slugify with multiple languages support.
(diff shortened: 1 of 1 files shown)


avelino:

- goreportcard.com: False
- coverage: False

spekary:
Oops. Hold up. I see there is a whole natural language category now. I will merge those in there.

Also, I see that some categories are using html headers rather than outlining for sub categories. I think that looks better so I am going to change to that.

spekary:
OK, should be good now. I reorganized the Natural Lanugage section too.

I removed https://github.com/nuance/go-nlp since it is officially archived and abandoned by the author.

spekary:
@avelino What do you think about this? Personally, I think giving better granularity by using subcategories makes the list more useful. I would like to continue going through each category looking for abandoned projects, but while I am there, if I can offer better organization, I would like to do that too. I don't want to proceed until I know whether this is acceptable.

avelino (approved):
LGTM, I like your proposal

> sorry for the delay in reviewing the contribution

PR #4105, opened 2022-03-15
Removing Go-Mix:

spekary:
- Last commit was over 2 years ago
- There are a number of open issues that are over 5 years old
- Does not have a link to a GoReportCard report
- Does not have link to a code coverage report

@charneykaye, @vulpivia, @haswalt, @kikht. 
Go-mix is scheduled to be removed from awesome-go by March 21, 2022 for the above mentioned reasons. If you would like to keep go-mix on awesome-go, please reply with updates to the repo.


avelino:

- goreportcard.com: False
- coverage: False

charneykaye:
Hiya team, thanks for taking care of this, and heartily sorry we didn't make the deadline. We're updating the go-mix library and will apply for re-entry after all requirements are met.

PR #4122, opened 2022-03-23
Renaming *Generation and Generics* category now that Go has native generics

spekary:
Renaming the Generation and Generics category to just "Generators" since that seems to describe the category better, and Generics is now a native part of the language, and so it will move into many different types of libraries.


avelino:

- goreportcard.com: False
- coverage: False

PR #4124, opened 2022-03-23
Removed encoding:

spekary:
- Last commit was 4 years ago
- There is a recent bug report that shows its not working with recent versions of Go, and there is no response from the author.
- It does not conform to current awesome-go standards

@zhenji, @zentures, encoding is scheduled to be removed on March 6, 2022, due to the above issues. If you would like to keep encdoing on awesome-go, please reply to this request with updates to the software.


avelino:

- goreportcard.com: False
- coverage: False

PR #4169, opened 2022-04-09
Removing go-cache

spekary:
Removing go-cache:

- No development for 3 years
- Does not conform to current awesome-go standards


avelino:

- goreportcard.com: False
- coverage: False

PR #4195, opened 2022-04-22
Removing session gate

spekary:
Removing sessiongate:

- Last commit was 4 years ago
- Current build is not passing its own tests

@f0rmiga , sessiongate is scheduled to be removed from awesome-go on May 6, 2022. If you would like to keep it here, please fix your repository to that it is passing tests with a current version of Go and shows code coverage of 80% or more, and then post here so we can confirm.


avelino:

- goreportcard.com: False
- coverage: False

Furumob39 (approved):

PR #4265, opened 2022-06-10
run stale repo check as test instead

### LINKED ISSUE #4264: Two declaration of `main()` in the `main` package
There are two declarations of `main()` in the same package `main` & my compiler is giving me errors. Is this duplicate expected?

https://github.com/avelino/awesome-go/blob/6c5cc61945a8a8555f0a81b5a814ace42fc4bbff/make_site.go#L29-L31
https://github.com/avelino/awesome-go/blob/6c5cc61945a8a8555f0a81b5a814ace42fc4bbff/test_stale_repositories.go#L273-L275

cglotr:
I think this fixes #4264. Instead of running the stale via `go run`, we can run it via `go test`.

I have run the workflow on my fork & it looks like it can work https://github.com/cglotr/awesome-go/actions.

What do you think @avelino?

### DIFF
--- a/stale_repositories_test.go
+++ b/stale_repositories_test.go
@@ -2,6 +2,7 @@ package main
 
 import (
 	"bytes"
+	"context"
 	"encoding/json"
 	"fmt"
 	"log"
@@ -9,6 +10,7 @@ import (
 	"os"
 	"regexp"
 	"strings"
+	"testing"
 	"text/template"
 	"time"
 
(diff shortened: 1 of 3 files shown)


avelino:

- goreportcard.com: False
- coverage: False

stale_repositories_test.go
```diff
@@ -232,7 +229,7 @@ func testStaleRepository() {
 		tokenSource := &tokenSource{
 			AccessToken: oauth,
 		}
-		client = oauth2.NewClient(oauth2.NoContext, tokenSource)
+		client = oauth2.NewClient(context.Background(), tokenSource)
```
phanirithvij, line 232: https://pkg.go.dev/golang.org/x/oauth2#pkg-variables
docs say
> Deprecated: Use context.Background() or context.TODO() instead.

PR #4280, opened 2022-06-14
Library go-email-normalizer moved to the correct category

dimuska139:
Rationale: go-email-normalizer is not implementing any of authentications schemes. So it can't be in "Authentication and OAuth" section


avelino:

- goreportcard.com: False
- coverage: False

PR #4580, opened 2022-10-27
refactor(tests): Use net/http constants for status codes

jidicula:
It's best to use the built-in library constants for status codes when possible as it improves readability, rather than having to look up status code definitions when otherwise changing behaviour.

---

Since this is a refactor to the repo's tests itself, I've removed the usual PR template that's meant for entries to the list.


avelino:

- goreportcard.com: False
- coverage: False

jidicula:
@avelino this is also a Hacktoberfest PR - if it passes muster, could you add the `hacktoberfest-accepted` label to this PR?

PR #4819, opened 2023-03-31
@golang_discussions

xojoc:


avelino (changes requested):
we have a rule not to create a new category with 1 item, you need at least 3 items, see more [here](https://github.com/avelino/awesome-go/blob/main/CONTRIBUTING.md)

PR #5064, opened 2023-10-17
Update README.md Added pinch of meaning to the readme

masterrmind1:
Corrected.

### DIFF
--- a/README.md
+++ b/README.md
@@ -54,7 +54,7 @@ _Special thanks to_
 
 **Contributing:**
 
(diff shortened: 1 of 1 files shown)


phanirithvij:
Any reason to prefer glance over gander? If simplifying and making it easier to read is the reason, I suggest "Please read the contribution guidlines" instead of taking a look/gander/glance.

skydiver1980 (approved):

PR #5093, opened 2023-11-05
add go-test-coverage service link to PR template

vladopajic:
this PR adds link to [go-test-coverage](https://github.com/vladopajic/go-test-coverage) in pull request template.

reasoning to include `go-test-coverage` are following:
- it provides the same level of functionality needed by repository owners (automated coverage check with badge display)
- it's free and open source
- it's easier to setup (no registration or permissions required)

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -5,7 +5,7 @@
 - repo link (github.com, gitlab.com, etc):
 - pkg.go.dev:
 - goreportcard.com:
-- coverage service link ([codecov](https://codecov.io/), [coveralls](https://coveralls.io/), etc.):
+- coverage service link ([go-test-coverage](https://github.com/vladopajic/go-test-coverage), [codecov](https://codecov.io/), [coveralls](https://coveralls.io/), etc.):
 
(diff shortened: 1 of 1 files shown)


vladopajic:
hello @avelino, could you review this PR? thanks!

skydiver1980 (approved):

phanirithvij:
I am not sure about this. Many repositories don't use GitHub or GitHub actions and thus suggesting it to everyone isn't ideal.
What do you think @avelino

vladopajic:
it's just additional option and developer can choose any method they would prefer.

vladopajic:
@phanirithvij @avelino 

here is some statistics: on randomly selected repositories (n=100) from awesome-go list 
- 100 (all) repositories are using github 
- 36 repositories don't use github workflow 
  - 10 repositories don't have continuous integration at all 
  
based on this statistics it looks like there are more repositories (~64%) that would already easily integrate with this tool (since those already use github workflow). 
from another view, 100% of those repositories are hosted on github, using github workflow is as easy as adding file in repository root.

since users can choose any method they like, there isn't any problem is suggesting any method.

personally, it's much more uncomfortable to use SaaS solutions that are already listed.

xhd2015:
not LGTM, we should not suggest using any specific tool

PR #5127, opened 2023-12-12
Fix stale repos test

alexbozhenko:
Fix the usage of the stale repo test. It has not been running since March 2023. Looks like this one is the last issue that was created by it: https://github.com/avelino/awesome-go/issues/4809 

See discussion at https://github.com/avelino/awesome-go/discussions/4871#discussioncomment-7010044
Currently, the stale_repository_test is invoked incorrectly. It does not accept the file name:
From `go help test`
```
usage: go test [build/test flags] [packages] [build/test flags & test binary flags]

'Go test' automates testing the packages named by the import paths.
```

If we look at the history of deletions, we will see that stuff is mostly added, and very occasionally deleted from this repo.
```
git log --pretty=format:"%h - %ad - %s" --shortstat -- README.md | grep -B1 'deletion' 
```
That results in lots of links to abandonware, which is not awesome at all.
Having that this website is mentioned at the very beginning of https://go.dev/wiki/, can we make it better?

## Reproducible test plan:
1. `export OAUTH_TOKEN=YOUR_TOKEN_FROM_GITHUB`
2. ```go test -timeout 99999s -v -run ^TestStaleRepository$``` 
(for this run I removed the [limit](https://github.com/avelino/awesome-go/blob/6051ff629e41337a6f125fe2269fc0e99074d39c/stale_repositories_test.go#L50) on number of stale repos, which is currently set to 10. Also, this will trigger github API ratelimit, which is 5000 requests per hour. I had to add delay between the calls.)
3. Check the issue it created: https://github.com/avelino/awesome-go/issues/5136
Note that out of 2413 github links in readme.md, 1217 were flagged as stale.


avelino (approved):
LGTM

PR #5192, opened 2024-02-02
fix: bexp repo url in readme

mkungla:
Update pkg link in readme to point to correct repo location


cassiobotaro:
Thanks @mkungla .

PR #5303, opened 2024-05-07
Use embed.FS for templates

utgwkk:
There is a TODO comment in main.go.

https://github.com/avelino/awesome-go/blob/18c06bce29392d446b0373d4b9f933419af51e15/main.go#L45-L49

So I implemented it by using `go:embed` directive and `embed.FS`

### DIFF
--- a/main.go
+++ b/main.go
@@ -3,6 +3,7 @@ package main
 
 import (
 	"bytes"
+	"embed"
 	"errors"
 	"fmt"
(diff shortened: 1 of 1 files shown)


ccoVeille:
Interesting idea you got, I never looked at [`embed.FS`](https://pkg.go.dev/embed#FS) thanks for sharing

PR #5312, opened 2024-05-20
chore: track code coverage in PRs

kilianc:
It would be nice to track code coverage for this repo

## Summary by CodeRabbit

- **Chores**
  - Updated permissions for pull requests to enhance workflow automation.
  - Modified test command to include coverage profiling for better test insights.
  - Added installation of Vim for improved editing capabilities in the workflow.
  - Integrated step for generating beautiful HTML coverage reports to visualize test coverage.

### DIFF
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -7,7 +7,8 @@ on:
   pull_request:
 
 permissions:
-  contents: read  #  to fetch code (actions/checkout)
+  pull-requests: write   # required for posting comments
+  contents: write        # required for git push
 
 jobs:
   build:
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -7,7 +7,7 @@ on:
   pull_request:
 
 permissions:
-  pull-requests: write   # required for posting comments
+  pull-requests: write   # required for posting comments (test)
   contents: write        # required for git push
 
 jobs:
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -20,5 +20,7 @@ jobs:
       run: go get -v -t -d ./...
     - name: Run tests
       run: go test -coverprofile=cover.out main_test.go main.go
+    - name: Install Vim
+      run: apt-get update; apt-get install -y vim;
     - name: Go Beautiful HTML Coverage
       uses: gha-common/go-beautiful-html-coverage@v1
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -27,7 +27,6 @@ jobs:
       if: steps.verify-changed-files.outputs.files_changed == 'true'
       run: go get -v -t -d ./...
     - name: Run tests
-      if: steps.verify-changed-files.outputs.files_changed == 'true'
       run: go test -coverprofile=cover.out main_test.go main.go
     - name: Install Vim
       if: steps.verify-changed-files.outputs.files_changed == 'true'
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -17,11 +17,20 @@ jobs:
     container: golang:latest
     steps:
     - uses: actions/checkout@v3
+    - name: Verify Changed files
+      id: verify-changed-files
+      uses: tj-actions/verify-changed-files@v20
+      with:
+        files: "!*.md"
     - name: Get dependencies
+      if: steps.verify-changed-files.outputs.files_changed == 'true'
       run: go get -v -t -d ./...
     - name: Run tests
+      if: steps.verify-changed-files.outputs.files_changed == 'true'
       run: go test -coverprofile=cover.out main_test.go main.go
     - name: Install Vim
+      if: steps.verify-changed-files.outputs.files_changed == 'true'
       run: apt-get update; apt-get install -y vim;
     - name: Go Beautiful HTML Coverage
+      if: steps.verify-changed-files.outputs.files_changed == 'true'
       uses: gha-common/go-beautiful-html-coverage@v1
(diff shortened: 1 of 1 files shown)


.github/workflows/tests.yaml
```diff
@@ -19,4 +19,8 @@ jobs:
     - name: Get dependencies
       run: go get -v -t -d ./...
     - name: Run tests
-      run: go test main_test.go main.go
+      run: go test -coverprofile=cover.out main_test.go main.go
+    - name: Install Vim
+      run: apt-get update; apt-get install -y vim;
```
phanirithvij, line 24: Why vim?

.github/workflows/tests.yaml
```diff
@@ -19,4 +19,8 @@ jobs:
...
+    - name: Go Beautiful HTML Coverage
+      uses: gha-common/go-beautiful-html-coverage@v1
```
phanirithvij, line 26: When trying to push to the cover branch github actions needs the write permission which only @avelino has access to give.

phanirithvij (changes requested):
I think this is okay, but @avelino needs to approve.

.github/workflows/tests.yaml
kilianc: it's required by the action, it usually comes with the os but the ci worker is a container with almost nothing in it :)

the action uses `ex` to replace some text from go's default html coverage

phanirithvij: Ok

.github/workflows/tests.yaml
avelino: by default the branch is `cover`, I believe we need to bring the current branch for it to work - we don't have the `cover` branch for this flow.

```
    # The branch to checkout or create and push coverage to.
    # Default: 'cover'
    branch: ''
```

avelino (changes requested):

avelino:
> I think this is okay, but @avelino needs to approve.

**general review:** I like the idea of putting this action

.github/workflows/tests.yaml
phanirithvij: I tested on my fork https://github.com/phanirithvij/awesome-go/pull/1
I had to enable github pages.

Now thinking about it, we already have github pages enabled for this repo, but can we enable one more github pages for the `cover` branch?

kilianc: > I believe we need to bring the current branch for it to work - we don't have the cover branch for this flow.

The action creates the branch for you. Letting the action create the branch and do a first time setup is the recommended way to do it. This is my suggestion:

- Let the action run in this PR, it will run all the setup code and create the branch.
- Turn on github pages and target `cover`
- Test that the links work and everything is kosher
- Merge!

kilianc: >Now thinking about it, we already have github pages enabled for this repo, but can we enable one more github pages for the cover branch?

if you already have github pages enabled for this repo, it complicates things. Let me think about it.

avelino: GitHub Pages is active, but we don't use it to host the site - it's active because it was used in the past, today we don't use it

I just changed the default branch of GitHub Pages to the `cover` branch and root folder
Screenshot 2024-05-24 at 15 04 19

kilianc: > But our repo's contributions are rarely changing go code. @kilianc is it possible to detect only if go code changes and run this step?

Changes to the go files are not the only thing that affects the coverage, but maybe it's acceptable here. I configured [verify-changed-files](https://github.com/marketplace/actions/verify-changed-files#using-the-contains-function) to ignore all `.md` file changes because that's what most of the PRs here are on.

I kinda liked having it on because I took as an opportunity for folks to educate themselves about code coverage and copy this process in their own project, but I totally understand that there are already multiple automated comments on each PR.

.github/workflows/tests.yaml
```diff
@@ -16,7 +17,21 @@ jobs:
...
+      uses: tj-actions/verify-changed-files@v20
+      with:
+        files: |
+          !*.md
     - name: Get dependencies
+      if: steps.verify-changed-files.outputs.files_changed == 'true'
       run: go get -v -t -d ./...
     - name: Run tests
-      run: go test main_test.go main.go
+      if: steps.verify-changed-files.outputs.files_changed == 'true'
```
phanirithvij, line 30: We need to run tests still, only skip the coverage step please.

phanirithvij (changes requested):

.github/workflows/tests.yaml
kilianc: it looks like it's not working either way :(

kilianc, line 30:
```suggestion
```

.github/workflows/tests.yaml
```diff
@@ -16,7 +17,20 @@ jobs:
...
     steps:
     - uses: actions/checkout@v3
+    - name: Verify Changed files
+      id: verify-changed-files
+      uses: tj-actions/verify-changed-files@v20
+      with:
+        files: |
+          !*.md
     - name: Get dependencies
+      if: steps.verify-changed-files.outputs.files_changed == 'true'
```
kilianc, line 27:
```suggestion
```

.github/workflows/tests.yaml
phanirithvij: @kilianc can you enable github actions and github pages for your fork for cover branch and try and run it.

kilianc: done: https://github.com/kilianc/awesome-go/pull/1#issuecomment-2143534773

but the action to detect changes is borked, I am going to use git directly or find a new one if that's ok

.github/workflows/tests.yaml
ccoVeille: > the action uses ex to replace some text from go's default html coverage

Why `ex` and not  `sed`?

Also, if you really need `vim`, I would suggest using `vim-nox` package

kilianc:
Folks where did we leave things on this?

PR #5473, opened 2024-11-01
contributing guide: add 5-month maturity quality standard.

yassinebenaid:
This PR adds a new rule to contributing guide:

>Projects must have al least 5 months of history since their first commit

PR #5492, opened 2024-11-27
tests: add repository maturity test

yassinebenaid:
**What was changed ?**
 Added a new test to check repository maturity

**Why do we need it?**
We need to check that projects listed here are seriously maintained. As we discussed at: #5473 

**How it works**
When you run the test, it reads all links from the entire README file, and for each link, uses the [github api](https://docs.github.com/en/rest?apiVersion=2022-11-28) to fetch the first commit made 5 months before the current date. If no commits were found, the test fails. 

Only links to github repositories are checked.

**How to run test?**
```bash
GITHUB_API_TOKEN="ghp_XXXXXXXXXXXXXXXXXXXXXXX" go test . -run ^TestMaturity$
```

**Why do we need an API key?**
`Rate Limit`: authenticated requests can send more requests per hour, considering that this list holds +1000 links. Not using and API key isn't an option.
>This is what worked for me, you might have a better option, correct me if I'm mistaking.

--- 

This PR only adds the code needed to run the test, when and how we should run them still needs a discussion, as stated above, we run tests against the entire link set. I don't think we must run this test on new PR because it may fail just because of rate limiting. 

I wonder if there is a way to run them only against links added in the PR.

### DIFF
--- a/maturity_test.go
+++ b/maturity_test.go
@@ -1,0 +1,77 @@
+package main
+
+import (
+	"encoding/json"
+	"fmt"
+	"net/http"
+	"os"
+	"strings"
+	"testing"
+	"time"
+
+	"github.com/PuerkitoBio/goquery"
+)
+
+var (
+	githubApiAuthorizationToken = os.Getenv("GITHUB_API_TOKEN")
+	minimumMaturityDate         = time.Now().AddDate(0, -5, 0)
+)
+
+func TestMaturity(t *testing.T) {
+	doc := goqueryFromReadme(t)
+	doc.Find("body li > a:first-child").Each(func(_ int, s *goquery.Selection) {
+		t.Run(s.Text(), func(t *testing.T) {
+			href, ok := s.Attr("href")
+			if !ok {
+				t.Error("expected to have href")
+			}
+
+			matches := reGithubRepo.FindStringSubmatch(href)
+			if matches == nil {
+				return
+			}
+
+			if len(matches) != 3 {
+				t.Fatalf("failed to extract repo and user from: %s, got [%v]", href, strings.Join(matches, ", "))
+			}
+
+			if err := checkRepositoryMaturity(matches[1], matches[2]); err != nil {
+				t.Fatal(err)
+			}
+		})
+	})
+}
+
+func checkRepositoryMaturity(user, repo string) error {
+	until := minimumMaturityDate.Format(time.RFC3339)
+	url := fmt.Sprintf("https://api.github.com/repos/%s/%s/commits?per_page=1&until=%s", user, repo, until)
+	request, err := http.NewRequest(http.MethodGet, url, nil)
+	if err != nil {
+		return fmt.Errorf("failed to create request for `%s`, %v", url, err)
+	}
+
+	request.Header.Set("Accept", "application/vnd.github+json")
+	request.Header.Set("X-GitHub-Api-Version", "2022-11-28")
+	request.Header.Set("User-Agent", "avelino")
+	request.Header.Set("Authorization", "Bearer "+githubApiAuthorizationToken)
+
+	http.DefaultClient.Timeout = time.Minute
+	httpRes, err := http.DefaultClient.Do(request)
+	if err != nil {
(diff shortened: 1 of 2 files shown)


maturity_test.go
```diff
@@ -0,0 +1,77 @@
...
+		return fmt.Errorf("failed to create request for `%s`, %v", url, err)
+	}
+
+	request.Header.Set("Accept", "application/vnd.github+json")
+	request.Header.Set("X-GitHub-Api-Version", "2022-11-28")
+	request.Header.Set("User-Agent", "avelino")
+	request.Header.Set("Authorization", "Bearer "+githubApiAuthorizationToken)
+
+	http.DefaultClient.Timeout = time.Minute
+	httpRes, err := http.DefaultClient.Do(request)
```
ccoVeille, line 59: I would have created a dedicated client, and not playing with the default one

maturity_test.go
```diff
@@ -0,0 +1,77 @@
...
+	until := minimumMaturityDate.Format(time.RFC3339)
+	url := fmt.Sprintf("https://api.github.com/repos/%s/%s/commits?per_page=1&until=%s", user, repo, until)
+	request, err := http.NewRequest(http.MethodGet, url, nil)
+	if err != nil {
+		return fmt.Errorf("failed to create request for `%s`, %v", url, err)
+	}
+
+	request.Header.Set("Accept", "application/vnd.github+json")
+	request.Header.Set("X-GitHub-Api-Version", "2022-11-28")
+	request.Header.Set("User-Agent", "avelino")
```
ccoVeille, line 55: The user agent could be a env variable too. And you should have a default one, one that respect the UserAgent format

maturity_test.go
```diff
@@ -0,0 +1,77 @@
...
+	"os"
+	"strings"
+	"testing"
+	"time"
+
+	"github.com/PuerkitoBio/goquery"
+)
+
+var (
+	githubApiAuthorizationToken = os.Getenv("GITHUB_API_TOKEN")
```
ccoVeille, line 16:
```suggestion
	githubAPIAuthorizationToken = os.Getenv("GITHUB_API_TOKEN")
```

Or to respect the GitHub syntax 


```suggestion
	apiTokenGitHub = os.Getenv("GITHUB_API_TOKEN")
```

maturity_test.go
```diff
@@ -0,0 +1,77 @@
...
+			matches := reGithubRepo.FindStringSubmatch(href)
+			if matches == nil {
+				return
+			}
+
+			if len(matches) != 3 {
+				t.Fatalf("failed to extract repo and user from: %s, got [%v]", href, strings.Join(matches, ", "))
+			}
+
+			if err := checkRepositoryMaturity(matches[1], matches[2]); err != nil {
```
ccoVeille, line 38: Ypu could use meaningful variable

```suggestion
			user, repo := matches[1], matches[2]
			if err := checkRepositoryMaturity(user, repo); err != nil {
```

maturity_test.go
```diff
@@ -0,0 +1,77 @@
...
+
+	request.Header.Set("Accept", "application/vnd.github+json")
+	request.Header.Set("X-GitHub-Api-Version", "2022-11-28")
+	request.Header.Set("User-Agent", "avelino")
+	request.Header.Set("Authorization", "Bearer "+githubApiAuthorizationToken)
+
+	http.DefaultClient.Timeout = time.Minute
+	httpRes, err := http.DefaultClient.Do(request)
+	if err != nil {
+		return fmt.Errorf("failed to fetch commits for [%s/%s], %v", user, repo, err)
```
ccoVeille, line 61: When something is repeated like %s/%s + user, repo in almost every error you return, plus the fact the user and repository are the argument of the function, I tend to do not add them in the error message. The caller should do it

Alexizx00789999 (approved):

ccoVeille (changes requested):
As you liked my feedbacks by dropping emoji on it, I will assume you will work on it when you have time. But for now, I prefer to mark as "request changes"

PR #5525, opened 2024-12-16
remove: https://github.com/mjibson/esc

yassinebenaid:
Remove: https://github.com/mjibson/esc (archived)


phanirithvij:
the whole resource embedding category can go ideally. but for now let's keep it simple.

yassinebenaid:
Agree, some of them are stale since 8 years. I'll open a separate PR for it.

PR #5573, opened 2025-01-22
remove a few archived repos

phanirithvij:
Ideally the historically useful among these could be moved to ARCHIVED.md or so.

But for now I don't have any such plans.


yassinebenaid (approved):
👍

PR #5729, opened 2025-07-02
feat: adding github.com/samber/hot - in memory caching lib

samber:
## We want to ensure high quality of the packages. Make sure that you've checked the boxes below before sending a pull request.

- [x] I have read the [Contribution Guidelines](https://github.com/avelino/awesome-go/blob/main/CONTRIBUTING.md#contribution-guidelines)
- [x] I have read the [Maintainers Note](https://github.com/avelino/awesome-go/blob/main/CONTRIBUTING.md#maintainers)
- [x] I have read the [Quality Standards](https://github.com/avelino/awesome-go/blob/main/CONTRIBUTING.md#quality-standards)

_Not every repository (project) will require every option, but most projects should. Check the Contribution Guidelines for details._

- [x] The repo documentation has a pkg.go.dev link.
- [x] The repo documentation has a coverage service link.
- [x] The repo documentation has a goreportcard link.
- [x] The repo has a version-numbered release and a go.mod file.
- [x] The repo has a continuous integration process that automatically runs tests that must pass before new pull requests are merged.
- [x] Continuous integration is used to attempt to catch issues prior to releasing this package to end-users.

## Please provide some links to your package to ease the review

- [x] forge link (github.com, gitlab.com, etc): https://github.com/samber/hot
- [x] pkg.go.dev: https://pkg.go.dev/github.com/samber/hot
- [x] goreportcard.com: https://goreportcard.com/report/github.com/samber/hot
- [x] coverage service link ([codecov](https://codecov.io/), [coveralls](https://coveralls.io/), etc.): https://app.codecov.io/gh/samber/hot/

## Pull Request content

- [x] The package has been added to the list in alphabetical order.
- [x] The package has an appropriate description with correct grammar.
- [x] As far as I know, the package has not been listed here before.

Thanks for your PR, you're awesome! :sunglasses:

## Summary by CodeRabbit

* **Documentation**
  * Added a new entry for the Go caching library "hot" to the "Caches" section, including a link and feature summary.

### DIFF
--- a/README.md
+++ b/README.md
@@ -701,6 +701,7 @@ _Data stores with expiring records, in-memory distributed data stores, or in-mem
 - [gocache](https://github.com/eko/gocache) - A complete Go cache library with multiple stores (memory, memcache, redis, ...), chainable, loadable, metrics cache and more.
 - [gocache](https://github.com/yuseferi/gocache) - A data race free Go ache library with high performance and auto pruge functionality
 - [groupcache](https://github.com/golang/groupcache) - Groupcache is a caching and cache-filling library, intended as a replacement for memcached in many cases.
+- [hot](https://github.com/samber/hot) - In-memory caching with expiration, jitter, background revalidation, thread-safety, sharding, metrics and much more.
 - [icache](https://github.com/mdaliyan/icache) - A High Performance, Generic, thread-safe, zero-dependency cache package.
 - [imcache](https://github.com/erni27/imcache) - A generic in-memory cache Go library. It supports expiration, sliding expiration, max entries limit, eviction callbacks and sharding.
 - [nscache](https://github.com/no-src/nscache) - A Go caching framework that supports multiple data source drivers.


ccoVeille:
Good design BTW @samber 

I have just read this yesterday

https://maypok86.github.io/otter/blog/cache-evolution/

BTW @maypok86 how does otter compete with `hot` ?

samber:
`hot` has more features ;)

maypok86:
> hot has more features ;)

Um, what kind? After a quick look, it doesn’t have any particular advantages. Just a lot of not very common methods.

maypok86:
So far, it looks like a sturdyc with generic keys—I’ll take a deeper look a bit later.

maypok86:
Actually, it turned out to be a decent cache, but I still haven't found any significant advantages.

Here's some code review—might be useful.

1. Adding v0 to awesome-go with a promise of [breaking changes](https://github.com/samber/hot/tree/main?tab=readme-ov-file#-installation) is just bad. I'd even say it calls into question the entire addition.
2. **Nanosecond-precision timestamps**. This is good, but I'm not sure if it's worth mentioning in the README (usually not). Using [Linkname](https://github.com/samber/hot/blob/main/internal/time.go#L5) is [not great](https://github.com/golang/go/issues/67401). And it's unnecessary here—you could just use [time.Since(time.Now())](https://github.com/jackc/puddle/pull/35).
3. Having a Prometheus collector implementation is good. But it has issues. At the very least, you often don't know the metrics users actually need. And the saddest part: the package architecture is really poor. Why force users to drag the Prometheus [client](https://github.com/samber/hot/blob/main/go.mod) into their dependencies?
4. The dependency on `size` is also unclear, especially since you're not accounting for quite a bit of overhead. But overall, it's fine.

If we're talking specifically about cache features and comparing it to `otter`:

1. Yes, `otter` doesn't offer policy choices, but you don't need them. W-TinyLFU will [always](https://maypok86.github.io/otter/performance/hit-ratio/) have top-tier hit rates. Especially since most of the policies you've implemented and offer as options aren't resistant to cache attacks (e.g., scanning).
2. `hot` is just a map with a mutex, with all the consequences described in the article. To solve high contention, you suggest sharding, which again won't help much in real-world workloads. `otter` also lacks this feature, but it doesn't need it. Otter can dynamically tune most of its data structures. `otter` also doesn't require sharding the eviction policy, which is critical since sharding on small caches will obviously reduce hit rates.
3. The "caching missed entries" feature is unnecessary and just bloats the API, in my opinion. Users can just store a `bool` alongside the value and even better understand their app's logic.
4. In 2025, `Range` not implemented as an iterator looks odd (especially from the author of `lo`). Did the Go authors work in vain? :)
5. Your TTL approach is ok, but... The expiration policy runs in O(n). Many caches handle this in O(log(n)). `Ristretto`, `Theine`, and `Otter` can do it in O(1). Jitter is also something you could add to `otter` in 5 minutes without any issues. `otter` also lets you choose a [TTL reset strategy](https://github.com/maypok86/otter/blob/main/expiry_calculator.go#L23). This is useful, for example, with API token caching. So, jitter is useful, but you can easily add it to `otter` yourself—otherwise, it's just less efficient.
6. Making `EvictionReason` a string is just bad. It's mostly needed for comparisons. Usually, `int` is used for this (see `slog` for reference).
7. Why do we need `LoggerChain` (especially a fully sequential one)?
8. What’s the point of `WarmUp` in this form? I don’t really see how it helps users (especially given the requirement to return a `map`). Even if reading from a file, you’d still have to allocate and populate a map, right?
9. Concurrent `loading` and `invalidation` might lead to dirty data.
10. `otter` can also tune node (`item`) sizes based on user config, saving a lot of memory.
11. No cost/weight-based cache sizing. Useful when entries vary greatly in size.
12. I’m generally not a fan of APIs like `SetMissingManyWithTTL`, but that’s nitpicking. It’s just that `HotCache` already has more methods than `otter.Cache`, yet far fewer fundamentally distinct features. Or take `Describe(ch chan<- *prometheus.Desc)`, should that really be a `HotCache` method?
13. I’m not sure, but it seems `hot` doesn’t support refresh (asynchronous loading). If true, that’s a huge loss. It can reduce app latency by orders of magnitude.
14. The `Loader` doesn’t accept `context.Context` as a parameter, even though it’s often used to store useful data (for example, for logging and tracing).

There might be more, but I’m tired of writing.

maypok86:
Okay, `hot` uses the term "revalidate" instead of refresh. Fundamentally, the same issues exist here as with loading. Although it's worth noting that `otter` [also](https://github.com/maypok86/otter/blob/main/refresh_calculator.go#L23) allows for more flexible cache configuration in this aspect.

And yes, even splitting into 256 shards and using a simple policy doesn't fully resolve high contention scenarios. It's important to note that in this context, `otter` essentially operates as a single-shard cache.

```go
seed := maphash.MakeSeed()
client := hot.NewHotCache[K, V](hot.LRU, capacity).WithSharding(256, func(key K) uint64 {
    return maphash.Comparable(seed, key)
}).Build()
```

```
BenchmarkCache/zipf_otter_reads=100%,writes=0%-8         	162626131	         8.489 ns/op	 117800079 ops/s	       0 B/op	       0 allocs/op
BenchmarkCache/zipf_theine_reads=100%,writes=0%-8        	69921410	        14.93 ns/op	  66973125 ops/s	       0 B/op	       0 allocs/op
BenchmarkCache/zipf_ristretto_reads=100%,writes=0%-8     	58397240	        20.37 ns/op	  49099848 ops/s	       0 B/op	       0 allocs/op
BenchmarkCache/zipf_sturdyc_reads=100%,writes=0%-8       	28350362	        41.37 ns/op	  24171966 ops/s	       0 B/op	       0 allocs/op
BenchmarkCache/zipf_ccache_reads=100%,writes=0%-8        	21627986	        70.41 ns/op	  14203144 ops/s	      52 B/op	       2 allocs/op
BenchmarkCache/zipf_hot_reads=100%,writes=0%-8           	41113078	        29.21 ns/op	  34232699 ops/s	       0 B/op	       0 allocs/op
BenchmarkCache/zipf_gcache_reads=100%,writes=0%-8        	 3774086	       290.2 ns/op	   3446119 ops/s	      16 B/op	       1 allocs/op
BenchmarkCache/zipf_ttlcache_reads=100%,writes=0%-8      	 2964015	       385.1 ns/op	   2596843 ops/s	      48 B/op	       1 allocs/op
BenchmarkCache/zipf_golang-lru_reads=100%,writes=0%-8    	 4567257	       251.6 ns/op	   3974618 ops/s	       0 B/op	       0 allocs/op
BenchmarkCache/zipf_otter_reads=75%,writes=25%-8         	77133626	        15.28 ns/op	  65447299 ops/s	       2 B/op	       0 allocs/op
BenchmarkCache/zipf_theine_reads=75%,writes=25%-8        	38410011	        26.28 ns/op	  38050182 ops/s	       0 B/op	       0 allocs/op
BenchmarkCache/zipf_ristretto_reads=75%,writes=25%-8     	28822191	        40.90 ns/op	  24452449 ops/s	       7 B/op	       0 allocs/op
BenchmarkCache/zipf_sturdyc_reads=75%,writes=25%-8       	11127548	       108.7 ns/op	   9199252 ops/s	      10 B/op	       0 allocs/op
BenchmarkCache/zipf_ccache_reads=75%,writes=25%-8        	15359818	        80.23 ns/op	  12463539 ops/s	      52 B/op	       2 allocs/op
BenchmarkCache/zipf_hot_reads=75%,writes=25%-8           	39460920	        29.38 ns/op	  34034373 ops/s	      12 B/op	       0 allocs/op
BenchmarkCache/zipf_gcache_reads=75%,writes=25%-8        	 3562063	       333.4 ns/op	   2999177 ops/s	      19 B/op	       1 allocs/op
BenchmarkCache/zipf_ttlcache_reads=75%,writes=25%-8      	 2853128	       408.7 ns/op	   2446547 ops/s	      33 B/op	       0 allocs/op
BenchmarkCache/zipf_golang-lru_reads=75%,writes=25%-8    	 4571716	       246.6 ns/op	   4055902 ops/s	       0 B/op	       0 allocs/op

```

maypok86:
Um, this is basically a race condition.

```go
if item, ok := c.cache.Get(key); ok {
    if !item.isExpired(nowNano) {
        return item, item.shouldRevalidate(nowNano), true
    }

    ok := c.cache.Delete(key)
	...
}
```

ccoVeille:
That's the most outstanding and unexpected feedback I saw for a while.

I was like "hey guys, your two libs seems similar and cool"

samber:
Hey @avelino 

I consider the API stable, but I'm looking for more feedback until 2026 before releasing v1. I will open a new PR as soon as it's done. 👌

TBH, the comment above is mostly opinionated (and trolling?). The race condition mentioned above isn’t a technical bug, but a deliberate trade-off. We accept a small window where two concurrent jobs might have a cache miss, to prevent locking a mutex for too long. Alexey read a comment in the code without making an effort to understand the problem under the hood.

Thanks for your feedback.

PR #5753, opened 2025-08-01
fix table of contents

miparnisari:
In https://github.com/avelino/awesome-go/pull/5694/files I broke the table of contents 😬


phanirithvij:
Thank you for the quick fix, I forgot about toc as well.

PR #6023, opened 2026-02-03
Remove GopherChina from the events list

scholar7r:
GopherChina domain is down, and @gopherchina group seems no longer active in the past 2 years.


phanirithvij:
Than you

PR #6391, opened 2026-06-09
Add fort to Security

djadmin:
fort is a macOS security auditing CLI. It runs 16 checks covering FileVault, SIP, firewall, Gatekeeper, SSH hardening, and more, reports a pass/fail score, and can auto-fix issues where safe to do so.

- Single static binary, installable via Homebrew
- MIT licensed
- Written in Go

Forge link: https://github.com/djadmin/fort
pkg.go.dev: https://pkg.go.dev/github.com/djadmin/fort
goreportcard.com: https://goreportcard.com/report/github.com/djadmin/fort


djadmin:
The `Running test` failure appears to be a pre-existing issue on `main` and unrelated to this PR.

PR #6419, opened 2026-06-21
remove defunct golang-sql-benchmark repo

imnotyu:
### Reason for Removal
The `golang-sql-benchmark` repository and the author's account have been deleted (404 error).


phanirithvij:
Incase anyone is interested https://github.com/c4pt0r/golang-sql-benchmark, but otherwise I don't know how popular a resource this was so as the original repo is gone I guess no need for this.

PR #6590, opened 2026-08-16
Add chicha-isotope-map

matveynator:
## Required links

- [x] Forge link (github.com, gitlab.com, etc): https://github.com/matveynator/chicha-isotope-map
- [x] pkg.go.dev: https://pkg.go.dev/github.com/matveynator/chicha-isotope-map
- [x] goreportcard.com: https://goreportcard.com/report/github.com/matveynator/chicha-isotope-map
- [x] Coverage: https://coveralls.io/repos/github/matveynator/chicha-isotope-map/badge.svg?branch=main

## Pre-submission checklist

- [x] I have read the [Contribution Guidelines](https://github.com/avelino/awesome-go/blob/main/CONTRIBUTING.md#contribution-guidelines)
- [x] I have read the [Quality Standards](https://github.com/avelino/awesome-go/blob/main/CONTRIBUTING.md#quality-standards)

## Repository requirements

- [x] The repo has a `go.mod` file and at least one SemVer release (`vX.Y.Z`).
- [x] The repo has an open source license.
- [x] The repo documentation has a pkg.go.dev link.
- [x] The repo documentation has a goreportcard link (grade A- or better).
- [x] The repo documentation has a coverage service link.
- [x] The repo has a continuous integration process (GitHub Actions, etc.).
- [x] CI runs tests that must pass before merging.

## Pull Request content

- [x] This PR adds/removes/changes **only one** package.
- [x] The package has been added in **alphabetical order**.
- [x] The link text is the **exact project name**.
- [x] The description is clear, concise, non-promotional, and **ends with a period**.
- [x] The link in README.md matches the forge link above.

## Category quality

- [ ] The packages around my addition still meet the Quality Standards.

PR #6591, opened 2026-08-16
Fix Security section alphabetical order

matveynator:
Fix the pre-existing `TestAlpha` failure on `main` by restoring alphabetical order in the Security section: `acme-proxy` must precede `acmetool`.

This maintenance fix is intentionally separate from #6590 so that the package submission continues to change exactly one item.

Validation: `go test main_test.go main.go`

### DIFF
--- a/README.md
+++ b/README.md
@@ -2420,8 +2420,8 @@ _Libraries for scientific computing and data analyzing._
 
 _Libraries that are used to help make your application more secure._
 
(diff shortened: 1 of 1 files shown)


matveynator:
CI note: `Running test` passes. `Post quality report` fails because the workflow treats every README-only maintenance move as a new package submission and requires forge/pkg/report links. This PR adds or changes no package; it only moves the existing `acmetool` line to repair the pre-existing `TestAlpha` failure on `main` and unblock #6590.

